PR #46, opened 2015-02-04
Bug 1105357 allow scheduling of tiles to publish

oyiptong:


splice/queries.py
```diff
@@ -358,6 +367,123 @@ def get_all_distributions(limit=100):
...
+    As a regular task, it is intended to run at least once hourly.
+    :minutes: amount of time in the past from the query time which is still viable
+    :dt_query: optionally set the date time to find schedules for
+    """
+    from splice.environment import Environment
+
+    env = Environment.instance()
+
+    if not minutes or not (0 < minutes < 60):
+        raise ValueError("minutes needs to be a number between 1..59 inclusive")
```
tspurway: should this be 0..59 inclusive?

splice/commands.py
```diff
@@ -265,6 +266,59 @@ def ingest_tiles(in_file, channel_id, out_path, console_out, deploy_flag, verbos
         traceback.print_exc()
 
 
+@DataCommand.option("-v", "--verbose", action="store_true", dest="verbose", help="turns on verbose mode", default=False, required=False)
+@DataCommand.option("-q", "--quiet", action="store_true", dest="quiet", help="turns on quiet mode", default=False, required=False)
+@DataCommand.option("-l", "--leniency-minutes", type=int, dest="leniency", help="Leniency period in minutes for scheduling", default=15)
```
tspurway: I would like a few more comments on 'leniency'.  It is not obvious why you need it, and it makes my inner 'distributed systems guy' feel queesy.

splice/queries.py
oyiptong: Stated as it is in the code, 1..59 is correct.

Arguably, a 0 time difference could be ok too

splice/commands.py
oyiptong: The reason why you need it is because clocks are not perfect, especially when running a scheduled job as a cron job.

The projected use case is to run a cron job every 15 minutes. Every scheduled distribution that is set within the past 15 minutes, up to the present minute, are candidates to be deployed.

If there were no leniency allowed, the command would only schedule jobs that have the exact same datetime as the time at which the command was run. Even a deviation of a second would prevent scheduled jobs from running.

There are more complicated methods to do this, e.g. rounding down the execution time to the nearest hour chunk. I felt that a leniency period was a simple way around this.

Either way, the cron job is set to run from only one host.

splice/queries.py
oyiptong: I remembered another factor why exact time difference was not implemented: the database implementations have a second resolution, whereas python has a microsecond resolution.

It's not a big deal, but it had me thinking of when an exact match would be useful... I couldn't find an occurrence for the current implementation

splice/static/html/partials/authoring.html
```diff
@@ -1,61 +1,159 @@
...
+                            </tr>
+                        </thead>
+                        <tbody>
+                            <tr>
+                                <td>{{ channelIndex[channelSelect.id].name }}</td>
+                                <td>{{ deployConfig.scheduled ? deployConfig.scheduled.toString() : "Now" }}</td>
+                            </tr>
+                            <tr ng-show="deployConfig.scheduled">
+                                <td></td>
+                                <td>{{ deployConfig.scheduled ? deployConfig.scheduled.toUTCString() : "" }}</td>
```
emtwo: Tested this out locally. It looks like in this line + line 21 above we get the date written twice in different formats if deployConfig.scheduled is true?

oyiptong: Yes that is correct. Once in local time and once in UTC so ed doesn't need to do mental gymnastics. Does it work as expected?

emtwo: yep, it works. Just couldn't tell if it was on purpose or not

oyiptong: Aha, yes. It was a feature request from Ed. I should probably make a comment.

splice/static/css/spliceApp.css
```diff
@@ -97,10 +97,48 @@ ol.tiles-list > li:before {
...
+}
+
+.modal-backdrop {
+    opacity: 0.7;
+}
+
+.modal-status-message {
+    word-wrap: break-word;
+    overflow-y: auto;
+    max-height: 700px;
```
emtwo: This may not be that important, but with this max-height, the modal doesn't fit on my screen completely.

splice/static/js/controllers/distribution.js
```diff
@@ -0,0 +1,91 @@
...
+      $scope.tiles = {};
+    }
+    return results.valid;
+  };
+
+  $scope.init = function(initData) {
+    $scope.payloadSchema = initData.schema;
+    $scope.setupChannels(initData.chans);
+    $scope.setupDistributions(initData.dists);
+    $scope.channelSelect = initData.chans[0];
```
emtwo: Just paranoid seeing indexing [0] here. Maybe a comment or check first in case `initData.chans` is empty?

splice/static/js/controllers/distribution.js
```diff
@@ -0,0 +1,91 @@
...
+        if (oldId == $scope.channels[i].id) {
+          $scope.channelSelect = $scope.channels[i];
+          break;
+        }
+      }
+      // note that channelSelect could still be null after this
+    }
+
+    if ($scope.channelSelect == null) {
+      $scope.channelSelect = $scope.channels[0];
```
emtwo: Same here about [0] indexing, I suppose there's no scenario where `$scope.channels` would be empty?

splice/static/js/controllers/upcoming.js
```diff
@@ -0,0 +1,88 @@
...
+  };
+
+  $scope.openRemoteDistribution = function(url) {
+    /**
+     * Open a distribution file given a URL
+     */
+    var chanId = $scope.channelSelect;
+    var cacheKey = chanId + url;
+
+    if (!$scope.cache.hasOwnProperty(cacheKey)) {
```
emtwo: This entire if/else statement seems awfully similar to the one in authoring.js in `$scope.$watch('versionSelect, ...)` Both are fetching and loading distribution json. Perhaps we can factor out some of this into a shared function?

tspurway:
@oyiptong r+

splice/static/js/controllers/distribution.js
oyiptong: There should always be channel data.

That said, I think you have a point. A more user-friendly error message will be shown in a coming commit.

splice/static/js/controllers/upcoming.js
oyiptong: hmm. good point. I missed that

splice/static/css/spliceApp.css
oyiptong: This is as designed. I do not think anyone will want to view those links in that UI. There's a feature to easily copy/paste the content in the status message in case one wants to look at the data in depth

emtwo:
@oyiptong r+

PR #47, opened 2015-02-26
use rsplit to be more concise

relud:
when extracting module and class name from config string

@oyiptong or @tspurway


oyiptong:
r+

PR #48, opened 2015-03-24
Fix bug 1146028 - Old tiles being served with old tile_index on new deploy 

oyiptong:
r? @tspurway


oyiptong:
or r? @emtwo

tspurway:
looks good @oyiptong 

r+

PR #50, opened 2015-04-01
update authoring publish success UI with upload status

oyiptong:
r? @emtwo

### DIFF
--- a/splice/static/js/controllers/authoring.js
+++ b/splice/static/js/controllers/authoring.js
@@ -106,16 +106,14 @@ angular.module('spliceApp').controller('authoringController', function($controll
         var msg = '<ol>';
         for (var url of data.urls) {
 
-          var uploadStatus = null;
-          var uploadClass = null;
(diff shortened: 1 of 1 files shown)


oyiptong:
screen shot 2015-04-01 at 4 32 59 pm

splice/static/js/controllers/authoring.js
```diff
@@ -105,7 +105,18 @@ angular.module('spliceApp').controller('authoringController', function($controll
...
+
+          var uploadStatus = null;
+          var uploadClass = null;
+          if (url[1] == true) {
+            uploadStatus = "new";
+            uploadClass = "text-success";
+          }
+          else {
+            uploadStatus = "cached";
+            uploadClass = "text-muted";
```
emtwo: nit: you could move these two lines up to replace line 109/110 where you set them to null to save a couple of lines of code

emtwo:
r+

PR #54, opened 2015-04-23
fix the migration

relud:
r? @tspurway


relud:
also, this migration works in stage

mostlygeek:
R+ from me. :)

relud:
also, I verified that the assumption I made that tile id will match adgroups id is valid in prod and stage

tspurway:
wfm, good work @relud

tspurway:
r+

PR #59, opened 2015-04-30
Bug 1159975 - Fix bug in Splice 1.1.11 where locale distributions are getting many extra tiles

oyiptong:
tracked at https://bugzilla.mozilla.org/show_bug.cgi?id=1159975


tspurway:
r+ @oyiptong

PR #62, opened 2015-05-07
Bug 1161192 - Implement frequency cap in splice

oyiptong:
tracked at https://bugzilla.mozilla.org/show_bug.cgi?id=1161192

This feature is optimizable. I.e. however way we mutate it, we should have a way to track how it has performed.


tspurway:
@oyiptong are there db migrations in a separate branch?  could you submit those in a PR as well?

oyiptong:
I forgot about the migrations. D'oh.

It's coming soon

ncloudioj:
r+ @oyiptong

PR #63, opened 2015-05-08
Fraud Prevention: Reporting Migrations

tspurway:
r? @oyiptong 

database migrations for new tables: blacklisted_ips, blacklist_stats_daily


oyiptong:
can you please add the entries in models.py?

splice/models.py
```diff
@@ -62,6 +62,13 @@ class AdgroupSite(db.Model):
     created_at = db.Column(db.DateTime(), nullable=False, default=datetime.utcnow)
 
 
+class BlacklistedIP(db.Model):
+    __tablename__ = "blacklisted_ips"
+
+    ip = db.Column(db.String(1024), nullable=False, primary_key=True)
+    created_at = db.Column(db.DateTime(), nullable=False, default=datetime.utcnow)
```
oyiptong: the last_seen column is missing

splice/models.py
```diff
@@ -62,6 +62,13 @@ class AdgroupSite(db.Model):
     created_at = db.Column(db.DateTime(), nullable=False, default=datetime.utcnow)
 
 
+class BlacklistedIP(db.Model):
+    __tablename__ = "blacklisted_ips"
+
+    ip = db.Column(db.String(1024), nullable=False, primary_key=True)
```
oyiptong: The length should be 16 chars

splice/models.py
tspurway: d'oh!

splice/models.py
ncloudioj: Not necessarily to be limited by 16 chars. AFAIK, a lot of vendors also offer CRID IP ranges, i.e. 192.168.100.0/24, instead of giving a very wide list of consecutive IPs.

tspurway: Nice.  Thanks.  I will enwiden it (yes, that's a word).

tspurway:
I refactored these migrations to support the new style of doing fraud detection / IP blacklists (see https://github.com/tspurway/infernyx/pull/20)

r? @oyiptong

oyiptong:
r+

PR #65, opened 2015-05-13
Bug 1161201 accept input of negative adjacency inclusion

oyiptong:
tracked at: https://bugzilla.mozilla.org/show_bug.cgi?id=1161201


oyiptong:
r? @ncloudioj 

This PR is rebased on top of #62

ncloudioj:
Shouldn't we use two separate PRs for two features?

tests/base.py
```diff
@@ -32,10 +32,10 @@ def tile_values(fd):
 
         def adgroup_values(fd):
             for line in fd:
-                row = [el.decode('utf-8') for el in line.split(',')]
+                locale, check_inadjacency_str = [el.decode('utf-8') for el in line.split(',')]
                 yield dict(zip(
-                    ('locale',),
-                    row))
+                    ('locale', 'check_inadjacency'),
+                    (locale, True if check_inadjacency_str == 'true' else False)))
```
tspurway: nit:  
True if check_inadjacency_str == 'true' else False
is the same as
check_inadjacency_str == 'true'

oyiptong: Where is my mind? :-)

tspurway:
r+ @oyiptong

PR #68, opened 2015-05-21
Bug 1161196 add custom explanation

ncloudioj:
tracked at [1161196 - Add custom explanation](https://bugzilla.mozilla.org/show_bug.cgi?id=1161196)

This PR will add two fields to the "adgroup" table: "name" and "explanation", repectively.
Note that the "explanation" is a predefined template with 2 arguments as "%1$S" and "%2$S".


oyiptong:
@ncloudioj can you please add a test for input sanitization?

oyiptong:
nevermind, i see a test that does just that

oyiptong:
At: https://github.com/ncloudioj/splice/blob/bug-1161196-add-custom-explanation/tests/test_ingest.py#L235

Can you move this test in a separate function?

oyiptong:
r+ after the change

ncloudioj:
Yep, it deserves a separate test.

oyiptong:
Nan, can you please merge --squash?

PR #70, opened 2015-05-22
Allow explanation with optional replacements

ncloudioj:
This PR refers to [Bug 1167404](https://bugzilla.mozilla.org/show_bug.cgi?id=1167404) - Allow for custom explanation with optional replacement targets


oyiptong:
r+

PR #71, opened 2015-05-25
Bug 1161197 accept inputs of start stop times

oyiptong:
This PR will:
* allow for both relative and absolute start times for campaigns

tracked at: https://bugzilla.mozilla.org/show_bug.cgi?id=1161197


splice/ingest.py
```diff
@@ -119,6 +173,14 @@ def ingest_links(data, channel_id, *args, **kwargs):
 
     country_locales = sorted(data.keys())
 
+    def remove_unserializable_data(tile):
+        if 'time_limits' in tile:
+            limits = tile['time_limits']
```
ncloudioj:
```python
limits.pop('start_dt', None)
limits.pop('end_dt', None)
```
should be the same

ncloudioj:
r+ @oyiptong 
The added comment above is merely a minor one, and it's totally up to your own preference :)

splice/ingest.py
oyiptong: Awesome!

PR #73, opened 2015-06-10
Bug 1167761 reuse assets in ingestion

ncloudioj:
This PR refers to [Bug 1167761](https://bugzilla.mozilla.org/show_bug.cgi?id=1167761) - Implement asset re-use in ingestion format for splice


oyiptong:
can you please also test generate_artifacts and distribute?

ncloudioj:
Good catch. Will do that.

oyiptong:
Is it possible to create the distribution JSON with asset re-use as well?

That would save a lot of bandwidth upon re-publishing tiles

oyiptong:
This would require changes in how ingest_links works. The returned ingested_data should contain the tile_id's for each tile, but the assets should be kept separate from the tile data just as it is in the ingest format.

oyiptong:
All future distribution json files stored on S3 should be in the new compact format

splice/static/js/controllers/distribution.js
```diff
@@ -107,6 +117,14 @@ angular.module('spliceApp').controller('distributionController', function($scope
...
       for (var i = 0; i < tiles.length; i++) {
         var tile = tiles[i];
+
+        // populate the imageURI and enhancedImageURI if tile is in compact format
+        if (tile.hasOwnProperty("imageURI") && assets.hasOwnProperty(tile.imageURI)) {
+          tile.imageURI = assets[tile.imageURI];
+        }
+        if (tile.hasOwnProperty("enhancedImageURI") && assets.hasOwnProperty(tile.enhancedImageURI)) {
+          tile.enhancedImageURI = assets[tile.enhancedImageURI];
+        }
```
ncloudioj: @oyiptong on line 151 below, chanId seems to be an object, and "chanId + url" turns out be "[object Object]https:aws....com". I wonder did you mean "var chanId = $scope.channelSelect.id"?

oyiptong: Good catch!

splice/ingest.py
```diff
@@ -154,12 +57,36 @@ def slice_image_uri(image_uri):
...
+    def remove_unserializable_data(tile):
+        if 'time_limits' in tile:
+            limits = tile['time_limits']
+            limits.pop('start_dt', None)
+            limits.pop('end_dt', None)
+        return tile
 
+    def populate_image_uris(tile, assets):
+        try:
+            tile["imageURI"] = assets[tile["imageURI"]]
```
oyiptong: this could be made clearer with a variable assignment and a comment

e.g.
```python
def populate_image_uris(tile, assets):
    """Replace Tile imageURI ID's by image payloads inplace"""
    try:
        image_uri_id = tile["imageURI"]
        tile["imageURI"] = assets[image_uri_id]
        ...
```

same goes for `enhancedImageURI`

splice/ingest.py
```diff
@@ -154,12 +57,36 @@ def slice_image_uri(image_uri):
...
+            if "enhancedImageURI" in tile:
+                tile["enhancedImageURI"] = assets[tile["enhancedImageURI"]]
+        except KeyError as e:
+            command_logger.error("Failed to find key: {0} for enhanced/imageURI"
+                                 " when inserting {1}.".format(e, json.dumps(tile, sort_keys=True)))
```
oyiptong: Instead of the full JSON dump of the tile, can you just state the tile data without the base64-encoded image?
This code will allow for the b64 encoded string for "imageURI" in case the enhanced asset doesn't exist

splice/ingest.py
ncloudioj: make sense, will change it

splice/ingest.py
```diff
@@ -205,6 +129,8 @@ def remove_unserializable_data(tile):
                     if not env.is_test:
                         conn.execute("LOCK TABLE tiles; LOCK TABLE adgroups; LOCK TABLE adgroup_sites;")
 
+                    if is_compact and not populate_image_uris(t, assets):
+                        continue
```
oyiptong: Can you please add a TODO comment here and reference the bug below?

This is debt we'll need to tackle. I opened a bug here: [Bug 1169302 - Collect and return Tile ingestion errors](https://bugzilla.mozilla.org/show_bug.cgi?id=1169302)

splice/ingest.py
```diff
@@ -487,3 +413,59 @@ def distribute(data, channel_id, deploy, scheduled_dt=None):
             insert_distribution(url, channel_id, deploy, scheduled_dt)
 
     return distributed
+
+
+def compress_payload(payload):
+    """ Compress payload by re-using image uri and enhanced image uri. Note
```
oyiptong: How about:

Make an inplace modification to the payload that compresses it by creating an asset dictionary and referring to the assets by ID in the payload

splice/ingest.py
```diff
@@ -154,12 +57,36 @@ def slice_image_uri(image_uri):
 
 
 def ingest_links(data, channel_id, *args, **kwargs):
+    """ Obtain links, insert in data warehouse
+
+    Params:
+        data: JSON payload with multiple schema versions. For versions prior to
+        1.1.19, the JSON schema does NOT re-use data for imageURI and
```
oyiptong: prior to 1.1.21 (which is the targeted version this change will be at)

splice/schemas.py
```diff
@@ -0,0 +1,147 @@
+""" JSON schema definitions for splice
+
+* default schema:
+    the imageURI and enhancedImageURI are not re-used, Splice versions prior to
+    1.1.19 use this schema
```
oyiptong: prior to 1.1.21

splice/static/js/controllers/authoring.js
```diff
@@ -98,9 +98,73 @@ angular.module('spliceApp').controller('authoringController', function($controll
      * Send tiles to backend for publication.
      * Assumes data is correct.
      */
+    var cloneTile = function (tile) {
+      if (null === tile || "object" !== typeof tile) return tile;
+
+      var copy = tile.constructor();
+      for (var attr in tile) {
+        if (tile.hasOwnProperty(attr)) {
```
oyiptong: this check is redundant

splice/static/js/controllers/authoring.js
```diff
@@ -98,9 +98,73 @@ angular.module('spliceApp').controller('authoringController', function($controll
      * Send tiles to backend for publication.
      * Assumes data is correct.
      */
+    var cloneTile = function (tile) {
```
oyiptong: a better name for this function could be: `cloneTileNoAsset`

Another possible way to clone, with slight additional memory usage is to:
```javascript
var cloneTileNoAsset = function (tile) {
  var copy = JSON.parse(JSON.stringify(tile));
  ["imageURI", "enhancedImageURI"].map(function (attr){
    copy[attr] = undefined;
  });
  return copy;
}
```

ncloudioj: haha, that's smart!!!

splice/static/js/controllers/authoring.js
```diff
@@ -98,9 +98,73 @@ angular.module('spliceApp').controller('authoringController', function($controll
...
+       * compress the payload for publishing. Note that the tiles might be cached,
+       * therefore we create a cheap (all imageURIs won't be cloned) copy here
+       */
+      var copies = {};
+      var uri2id = {};
+      var assets = {};
+      var id = 0;
+
+      for (var locale in tiles) {
+        if (tiles.hasOwnProperty(locale)) {
```
oyiptong: this check is redundant

ncloudioj: really? I thought that was the standard way to iterate over an Object

oyiptong:
```javascript
for (var locale in tiles) {
...
}
```
This will iterate over the properties of the object `tiles` and store it in a local var `locale`.
The next line is a check whether the property is in the `tiles` object.

If the property `locale` wasn't in `tiles`, line 128 wouldn't  have executed.

Makes sense?

ncloudioj: That's true. I'd recommend to keep this check though. The "hasOwnProperty" excludes all the properties inherited from its prototype chain.

tests/test_ingest.py
```diff
@@ -695,7 +785,9 @@ def test_generate_artifacts(self):
 
         tile = fixture["STAR/en-US"][4]
 
-        data = ingest_links({"STAR/en-US": [tile]}, self.channels[0].id)
+        # making a copy for tiles because distribute will attempt to compress
+        # it in place. That will affect the following tests
+        data = ingest_links({"STAR/en-US": [copy.deepcopy(tile)]}, self.channels[0].id)
```
oyiptong: A better way to do this is perhaps to make a deep copy within ingest_links so that the original data is not mutated.
The original version also mutated the input data, and I think that was wrong.

splice/static/js/controllers/authoring.js
oyiptong: I think in this use of the data, we don't really care about prototypical inheritance.
Since the object is data (to be serialized) and not code, we won't ever care.

In fact, for this purpose, the aesthetics of this is better:
```javascript
if ("property" in obj) {
}
```

But either property checks are fine.

That said, I'd rather the code be more readable at this point, with less lines, i.e. only having 1 check

tests/test_ingest.py
ncloudioj: Yes, that bugs me for awhile. Let's clean this up by making a deep copy.

oyiptong: :+1:

splice/static/js/controllers/authoring.js
ncloudioj: Fair enough, will remove this check.

oyiptong:
r+

PR #74, opened 2015-06-10
Bug 1171116 - Add Firefox prerelease channel and add tracking per channel

oyiptong:
There should be an additional channel added, and it would be "desktop-prerelease"

Make desktop release channel a targetable feature.

i.e. it should create an entry a new adgroups column: 'channel'

That enables the ability to track channel performance separately.

The previous channel state can be obtained by unpacking and storing data from the distributions s3 links.
New tile_ids will probably not be necessary, since the only tiles that exist in another channel right now are the android ones.

more at https://bugzilla.mozilla.org/show_bug.cgi?id=1171116


ncloudioj:
r+ @oyiptong

PR #80, opened 2015-07-14
Upgrade alembic migration for multiple databases

ncloudioj:
This refers to Bug 1181659 - Create Alembic migrations for new RDS campaign database (postgres)


migrations/env.py
```diff
@@ -2,6 +2,10 @@
 from alembic import context
 from sqlalchemy import engine_from_config, pool
 from logging.config import fileConfig
+import logging
+import re
+
+USE_TWOPHASE = False
```
oyiptong: When is USE_TWOPHASE ever true?

ncloudioj: Good catch @oyiptong , I've also noticed that. We don't need the two phase commit here though. This option is from the Alembic template for the multi-db scenario, I didn't remove it mainly for the compatibility concern. I think it's OK here to keep all its default options. What do you think?

oyiptong: I think it's ok

oyiptong:
r+

PR #89, opened 2015-08-17
Initial commit 2.0

ncloudioj:
This PR includes following topics:

1. remove the deprecated modules and tests.
2. add Flask-Restful dependency.
3. upgrade Flask-Migrate to 1.5.0.


rlr:
Delete code is my favorite kind of code. I don't know enough about this yet to know if you deleted something you shouldn't, but I trust it's all good.

The flask restful stuff looks good. Thanks for the CSRF setting.

Flask migrate changes make sense.

r+  :shipit:

ncloudioj:
@rlr Thanks. We have to delete the whole old trafficking modules, because the upcoming model change is not compatible with the previous one. Otherwise, we'll need to maintain two different schemas in the database, which is pretty confusing and hard to maintain.

PR #91, opened 2015-08-18
Fix the broken alembic template

ncloudioj:
r? @rlr


rlr:
this fixes the issue for me. r+

PR #92, opened 2015-08-18
Add Restful APIs for adgroup

ncloudioj:
r? @rlr


splice/queries/adgroup.py
```diff
@@ -0,0 +1,82 @@
...
+def get_adgroups_by_campaign_id(campaign_id):
+    from splice.environment import Environment
+
+    env = Environment.instance()
+
+    rows = (
+        env.db.session
+        .query(Adgroup)
+        .filter(Adgroup.campaign_id == campaign_id)
+        .order_by(Adgroup.id.asc())
```
rlr: why asc over desc? does it make sense to have old ones first? I guess it doesn't matter if we are returning all. We can sort differently in UI

splice/web/api/adgroup.py
```diff
@@ -0,0 +1,130 @@
+from flask import Blueprint
+from flask_restful import Api, Resource, marshal, fields, reqparse
+
```
rlr: I usually only see one empty line when separating imports like this. Just what I'm used to. If this is the way this project does it, that's fine. Just making sure it wasn't an accident.

oyiptong: I think one line is better. But it's subjective and up to you. flake8 isn't complaining at the very least ;-)

splice/queries/adgroup.py
ncloudioj: That's correct. How to sort the adgroups is largely up to the UI logic.
Here we just return all the adgroups in ascending order, which has the same order as let the old ones go first.

oyiptong: desc may be better so we have less front-end code to write?

splice/web/api/adgroup.py
ncloudioj: This is an unnecessary empty line. It should be one line here. Will fix it.

tests/api/test_adgroup.py
```diff
@@ -0,0 +1,113 @@
...
+            response = self.client.get(url)
+            assert_equal(response.status_code, 200)
+            resp = json.loads(response.data)
+            assert_equal(len(resp["message"]), len(adgroups))
+
+    def test_post(self):
+        url = url_for('api.adgroup.adgroups', campaign_id=self.new_campaign_id)
+        data = json.dumps(self.new_adgroup)
+        response = self.client.post(url, data=data, content_type="application/json")
+        assert_equal(response.status_code, 201)
```
rlr: It would be good to verify that the adgroup is actually in the database.

tests/api/test_adgroup.py
```diff
@@ -0,0 +1,113 @@
...
+        new_adgroup = json.loads(response.data)["message"]
+
+        url = url_for('api.adgroup.adgroup', campaign_id=self.new_campaign_id, adgroup_id=new_adgroup["id"])
+        new_adgroup["type"] = "suggested"
+        data = json.dumps(new_adgroup)
+        response = self.client.put(url, data=data, content_type="application/json")
+        assert_equal(response.status_code, 400)
```
rlr: It isn't immediately clear to me why this is a 400. Maybe a doc string or comment would help.

tests/api/test_adgroup.py
```diff
@@ -0,0 +1,113 @@
+from nose.tools import assert_equal
+from flask import url_for, json
+from tests.base import BaseTestCase
+from tests.populate_database import parse_csv
+from collections import defaultdict
+
+
+class TestAdgroup(BaseTestCase):
```
rlr: Tests look good. I think a few comments and/or docstrings will make it easier to read them. One thing I think is missing is to check that stuff actually makes it to the database. The API could be lying.

splice/queries/adgroup.py
ncloudioj: sure, i can make the change so that the most recent created adgroups will show up first by default.
It really doesn't matter for now.

rlr:
I don't see any major issues. Just those minor things and nitpickings. :boom:

tests/api/test_adgroup.py
ncloudioj: Yay, will drop a comment for it. This is to test if an adgroup is of the "suggested" type, it must have a non-empty category string attached too.

tests/api/test_adgroup.py
ncloudioj: Good point. Will fill out the missing parts.

As you might have noticed that we've introduced the session scope for all the HTTP POST and PUT requests. If there is no exception captured there, all the changes in that session should be committed successfully. Will double check it's actually in the database in the tests. Be safe :)

ncloudioj:
Thanks for the feedback, folks.
Various changes sit here fc997d064f1318ff364c9db406dfb1cb9e12a256
r? @rlr

rlr:
yay! looks good. r+ :ship:

ncloudioj:
Close #92 cbe87d3084162143ff97d5399d6740515293ba66

PR #93, opened 2015-08-20
Add restful APIs for tiles

ncloudioj:
r? @rlr


splice/queries/tile.py
```diff
@@ -0,0 +1,74 @@
...
+        where(Tile.enhanced_image_uri == record["enhanced_image_uri"]).
+        where(Tile.type == record["type"]).
+        where(Tile.status == record["status"]).
+        where(Tile.explanation == record["explanation"]).
+        where(Tile.adgroup_id == record["adgroup_id"])).scalar()
+    return ret
+
+
+def insert_tile(session, record):
+    if not tile_exists(session, record):
```
rlr: so, instead of doing this extra query, I catch the IntegrityError. Saves a query. Easier to say sorry than ask permission ;)

rlr: It makes all that code above unnecessary. And keeps the constraint/integrity checks in the database.

splice/web/api/tile.py
```diff
@@ -0,0 +1,125 @@
...
+        try:
+            # validate background color, url, and image fields again
+            jsonschema.validate(args, API_TILE_SCHEMA)
+            with session_scope() as session:
+                new = insert_tile(session, args)
+        except jsonschema.exceptions.ValidationError as e:
+            return {"message": e.message}, 400
+        except InvalidRequestError as e:
+            return {"message": e.message}, 400
+        except Exception as e:
```
rlr: I missed this in the previous pull request. Catching Exception is not good. It will hide all kinds of errors and you aren't even logging it or doing anything. It's better just to catch the exceptions that we expect can be normal (like IntegrityError) and let the rest be handled where all rest of the unhandled app exceptions are handled. I assume they send emails or do something special so we can find out about them.

splice/queries/tile.py
ncloudioj: Yup, I'd like to do that. But in this case, we can't merely rely on the database to check the existence(or Integrity). For one thing, it doesn't have any unique field. Also each tile is supposed to be immutable. The only mutable field is its status. So if we insert two exact tiles without this check, it'll happily do that for us.

rlr: ah, this is something special with the immutable thing.. OK, still have to wrap my head around that.

splice/web/api/tile.py
ncloudioj: Yes, this is not sufficient. It swallows all the exceptions, and sometime even returns some confusing message back to the client. A better exception handling and logging are definitely needed here.

splice/queries/tile.py
ncloudioj: yup, i really want to delegate this check (as well as adgroup) to database, but the immutability adds some extra work for us. So we have to check that when inserting every new record.

splice/web/api/tile.py
```diff
@@ -0,0 +1,121 @@
...
+            with session_scope() as session:
+                tile = update_tile(session, tile_id, args)
+        except NoResultFound as e:
+            return {"message": e.message}, 404
+        else:
+            return {"message": marshal(tile, tile_fields)}, 200
+
+
+api.add_resource(TileListAPI, '/adgroups/<int:adgroup_id>/tiles', endpoint='tiles')
+api.add_resource(TileAPI, '/adgroups/<int:adgroup_id>/tile/<int:tile_id>', endpoint='tile')
```
rlr: see, I find it weird that if you know the tile_id then why do you also have to pass the adgroup_id? Why is it required and not the campaign and account as well then?

rlr:
I think we can land this and then make all the different API endpoints consistent one way or another. It's a cosmetic detail. Having it working matters more. r+

ncloudioj:
@rlr Agreed! Just dropped a comment in your PR, keep whatever we have for now. Let's discuss this in next week.

PR #95, opened 2015-08-21
Fix for Account API when all fields arent sent (PUT).

rlr:
Uncovered this little bug while testing the campaign API. The issue was that the request parser defaults missing values to None by default.

r? @ncloudioj


ncloudioj:
r+ @rlr  Good catch :+1:

PR #98, opened 2015-08-23
[bug 1189691][bug 1189693][bug 1189695] Campaign API

rlr:
r? @ncloudioj 

I want to add some more thorough tests updating more of the fields. I'll get to that on the airplane tomorrow and we can add after.

### DIFF
--- a/splice/web/api/campaign.py
+++ b/splice/web/api/campaign.py
@@ -123,7 +123,6 @@ class CampaignAPI(Resource):
                                        help='Status', location='json',
                                        store_missing=False)
 
-
         super(CampaignAPI, self).__init__()
 
     def get(self, campaign_id):
--- a/splice/web/api/campaign.py
+++ b/splice/web/api/campaign.py
@@ -1,0 +1,157 @@
+from datetime import datetime
+
+from flask import Blueprint, request
+from flask_restful import Api, Resource, marshal, fields, reqparse
+from sqlalchemy.exc import IntegrityError
+
+from sqlalchemy.orm.exc import NoResultFound
+from splice.queries.common import session_scope
+from splice.queries.campaign import (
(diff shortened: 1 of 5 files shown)


splice/web/api/campaign.py
```diff
@@ -0,0 +1,156 @@
...
+                campaign = update_campaign(session, campaign_id, args)
+        except NoResultFound as e:
+            return {'message': e.message}, 404
+        except IntegrityError as e:
+            return {'message': 'Invalid account_id or channel_id provided.'}, 400
+
+        return {'result': marshal(campaign, campaign_fields)}, 200
+
+
+api.add_resource(CampaignListAPI, '/campaigns', endpoint='campaigns')
```
ncloudioj: Rather than passing the account_id as an argument here, why not use '/account/<int:account_id>/campaigns' here? Is there any advantage by doing that, or do you have other concerns? For other models, we've decided to build the endpoint with all the IDs in the URL. So we'd better be consistent.

rlr: hmm. Well, I see that as a filter. I want to filter campaigns by account. In the future I might want to filter by more or different things. And filters usually go as query parameters in the URL.

That said, I don't care that much.

It seems to me if we were doing something like:

/accounts/123/campaigns

Then adgroups should really be:

/accounts/123/campaigns/123/adgroups

And so on. As you see, that gets ugly quick.

ncloudioj: "/accounts/123/campaigns/123/adgroups" is definitely a big **nope**. I've at least omitted the "/accouts/123/" part :). My preferred URL actually is "mozilla/directory/mdn/tile", because nobody can remember those IDs easily.

To be consistent, let's switch to the query string (as what you did here) then. I'll change that for my models in the next sprint.

ncloudioj:
@rlr r+

PR #99, opened 2015-08-26
bug-1181640-part2-generate-account-and-campaign-tables

tspurway:
r? @ncloudioj

This PR contains models, migrations and scripts for accounts, campaigns and countries tables/data


scripts/index_walker.py
```diff
@@ -0,0 +1,190 @@
...
+import json
+import datetime
+from collections import defaultdict
+from itertools import chain
+from splice.environment import Environment
+from sqlalchemy.orm import sessionmaker
+from sqlalchemy.sql import select, insert, update
+from sqlalchemy import create_engine
+from splice.models import Tile, Adgroup, Country, Account, Campaign, CampaignCountry
+from tld import get_tld
```
ncloudioj: This library needs to be added in `requirements.txt`

scripts/index_walker.py
```diff
@@ -0,0 +1,190 @@
...
+ARBITRARY_FUTURE = datetime.datetime.strptime('2525-01-01', "%Y-%m-%d").date()
+
+
+def main():
+    """
+    This script is going to populate Account and Campaign database structures.  It does this
+    by reading the currently deployed tile distributions (s3), where it determines the currently active
+    tile set, as well as the geo-targetting data (currently only country level) for each tile/adgroup.
+
+    The script also popultates/migrates data from the 'countries' table in the stats database.
```
ncloudioj: Load 'countries' from fixtures.

scripts/index_walker.py
```diff
@@ -0,0 +1,190 @@
...
+    * the TLD+1 of the target_url
+    * the locale of the adgroup
+    * the channel of the adgroup
+    * the 'active' flag (determined as explained above) of the adgroup
+
+    One campaign row will be assigned for each unique campaign detected.
+
+    The script will populate the adgroup.campaign_id with the campaign that the adgroup fits into.
+
+    Account objects are considered unique by considering the TLD+1 of the campaign.  All campaigns that
```
ncloudioj: For content tile partners `TLD+1` might not be sufficient. For example, Conde Nast has 21 sub-brands, which all equipped with different domains.

splice/models.py
```diff
@@ -6,12 +5,54 @@
 metadata = db.metadata
 
 
+class Account(db.Model):
+    __tablename__ = "accounts"
+
+    id = db.Column('id', db.Integer(), autoincrement=True, primary_key=True, info={"identity": [1, 1]})
+    name = db.Column('name', db.String(length=255), nullable=True)
```
ncloudioj: I think `name` should be mandatory.

ncloudioj:
As we aim to add `accounts` and `campaigns` to the current system, there are a few things we have to think about,

1. How to create new accounts and new campaigns? Adding another UI for Splice? Or auto-generating them through a script whenever we have new tiles added? 

2. How to link the account and campaign info to the tiles and ad groups? This is even harder than the first one. Currently, we generate tiles in a batch manner, and there is no campaign related info in the raw tile payload. Filling up the relation by executing the script might get annoying as there are constantly increasing new partners and campaigns coming to this game.

tspurway:
i just also realized that my optimization that disposes of all country data for campaigns that have a STAR is wrong.  We still need to populate the campaign_countries for this case and cannot simplify/optimize away the non-STAR countries.

tspurway:
We have decided to put this patch into splice-2.0 branch, and so our concerns about keeping production data up-to-date isn't an issue

PR #100, opened 2015-08-26
Account and Campaign API Fixes.

rlr:
* add store_missing=False to name in case of PUT for campaigns
* make account_id required when requesting campaign list

r? @ncloudioj

### DIFF
--- a/splice/web/api/campaign.py
+++ b/splice/web/api/campaign.py
@@ -80,7 +80,10 @@ campaign_parser.add_argument(
 
 class CampaignListAPI(Resource):
     def __init__(self):
-        self.reqparse = campaign_parser
+        self.reqparse_post = campaign_parser
+        self.reqparse_get = reqparse.RequestParser()
+        self.reqparse_get.add_argument(
+            'account_id', type=int, required=True, help='Account ID', location='args')
 
         super(CampaignListAPI, self).__init__()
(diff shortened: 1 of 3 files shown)


splice/web/api/campaign.py
```diff
@@ -89,7 +89,11 @@ def get(self):
 
         Takes an optional account_id as a query string argument.
         """
-        campaigns = get_campaigns(request.args.get('account_id'))
+        parser = reqparse.RequestParser()
```
ncloudioj: better create one single RP instance in `__init__` and reuse it. i know premature optimization is the root of evil :)

rlr: but the other method uses a different parser (for POST). That's why it's local here.

ncloudioj: I just created a self.reqparse_get for this. Totally up to you, not an issue at all.

ncloudioj:
r+ @rlr

splice/web/api/campaign.py
rlr: ah I see. I'll do that for consistency.

rlr:
Pull request 100!!!!!!! :tada: 
fff0a262ca4243a3218a024f6cc4259b76d3c0c3

PR #102, opened 2015-08-27
Refactor the return format

ncloudioj:
Using "results" for multiple return values, and "result" for single ones.

r? @rlr


rlr:
yay r+!

ncloudioj:
Thanks @rlr. Finally, we're on the same track now :beers:

PR #104, opened 2015-08-28
Remove table locks

ncloudioj:
The removal of table locks enables Splice to survive from the locking failures, e.g. some applications might hold a long-lived transactions while they're running.

**Edge Case**: When two Splice instances are ingesting a same tile at the exact same time, it may end up inserting the same tile into the database twice. That should be very rare, but it's still possible. Only allowing one Splice instance running at a time could eliminate this issue.


oyiptong:
Can we have nested transactions? That may take care of the edge case.

Furthermore, there would need to be code to handle transaction failures in the case that you mention.

Composite unique key will also help to avoid some duplication scenarios.

ncloudioj:
Adding composite unique keys as well as replacing plain SQL statements with SQLAlchemy operations should eliminate those edge cases now.

oyiptong:
SQLAlchemy version pinning worked huh?

ncloudioj:
Running CircleCI without cache will update the latest SQLAlchemy automatically. So no need to change the requirements.txt here.

oyiptong:
awesome!

splice/queries.py
```diff
@@ -1,37 +1,44 @@
...
+def session_scope(conn):
+    try:
+        sm = sessionmaker(bind=conn)
+        session = sm()
+        session._model_changes = {}  # this is a workaround to make Flask-SQLAlchemy happy
+        yield session
+        session.commit()
+    except:
+        session.rollback()
+        raise
```
oyiptong: We want to do some logging here

oyiptong: Actually, never mind. The error bubbles up and is logged later

splice/ingest.py
```diff
@@ -135,100 +135,86 @@ def populate_image_uris(tile, assets):
...
+                            log_msg = "NOOP: id:{0} already exists".format(f_tile_id)
+
+                        else:
+                            # Either f_tile_id was not provided or the id's provided differ
+                            t["directoryId"] = db_tile_id
+                            log_msg = "IGNORE: Tile already exists with id: {1}".format(f_tile_id, db_tile_id)
 
                 except Exception as e:
-                    trans.rollback()
                     command_logger.error("ERROR: {0}\nError inserting {1}.  ".format(e, json.dumps(
```
oyiptong: Can we bubble up the error so that the originating caller, e.g. the API handler can return an error to the end-user?

ncloudioj: Sure we could. But the original function doesn't pop up exceptions, which means it tolerates some individual failures during the tile ingestion. @oyiptong Did you want to change this to "all-or-nothing" scheme? i.e. rollback the whole transaction whenever seeing errors.

oyiptong:
all-or-nothing is good. it gives the opportunity to retry.

The upload burden is much smaller now, and the database insertion costs is minimal, and about to get even faster too.

oyiptong:
r+

ncloudioj:
Just realized that the composite unique key for a single table is not correct here. Checking uniqueness on a single table doesn't make any sense. For example, a single tile that runs in multiple channels. We can't reject to insert it without considering its channel_id, which is only available in the `adgroups` table. In addition, we've wrapped the whole tile/adgroup insertion into a single transaction, the `tile_exists` will guarantee the uniqueness of a new `(adgroup, tile)` pair. 

Makes sense? @oyiptong

oyiptong:
Nan, I don't think it will work without the unique key. While it's true that the key is not across tables, it prevents concurrent duplicates within the same table.

Remember the scenario of 2 concurrent inserts? The constraint within the `tiles` or `adgroups` table is what prevents them from both occurring.

Now, the insertion requires that both constraints are met. In the case of concurrent inserts, one or the other will fail, meaning potentially both transactions will fail, requiring the user to try again.

oyiptong:
You're right, the current constraint doesn't make sense. You could have the same tile for a different channel

ncloudioj:
@oyiptong I understand your concern. But the issue with single table unique constraint is that it'll reject some legit insertions in the single writer case.

For example, say we only want to change a tile's rollover image, and keep its other attributes the same. It'll pass the cross-table uniqueness check, however it'll fail the single-table check. Because for this tile, all the columns of `adgroups` remain the same. This is not uncommon in the current `adgroups` table, which has only 105 unique items out of 2763 entries.

Looks like to avoid concurrent inserts, we still have to rely on some locking mechanism, either from the database, or the single-writer PID file lock :). What do you think? @oyiptong

oyiptong:
I thought the tile unique key included the `adgroup_id`, but I guess I was wrong.
The tile unique key should include the `adgroup_id`.

oyiptong:
Yes, the idea is to use the constraint as a concurrency mechanism

oyiptong:
As in, the transaction gets rejected at commit time if the constraint isn't met.

ncloudioj:
As we've switched from "one-transaction-per-tile" to "all-or-nothing", it actually gets harder to prevent concurrent insertions happening without locks.

I wonder is "locking table + insert all in a single transaction" gonna help? Well, it'll still fail to acquire the lock if Zenko holds the read connection. But once it gets the lock, it'll work hard to finish the insertion as fast as possible. The difference is in this case it only tries to acquire the table locks once.

Thoughts? @oyiptong

oyiptong:
I think we can get rid of the locks. Wouldn't the above work with the constraints + all-or-nothing?

In the concurrent transaction scenario, one will fail at commit time.
`adgroup_id` in the composite unique key would take care of the same-tile in different channel scenario.

Can we write some tests?

ncloudioj:
Well, this bug sounds somewhat off-topic now, instead of entirely removing the table locks, we've just lowered down the frequency of table locking from lock-per-tile to lock-only-once.

Not only should it speed up the single insert, but also it avoids the multiple concurrent inserts. As it significantly eliminates some locking overheads, given the current tile set (400+ tiles), the benchmark has showed some promising results.

r? @oyiptong, one more round please :), thank you!

oyiptong:
r+

PR #105, opened 2015-09-03
bug-1197962-add-adgroup-categories

tspurway:
This PR adds the adgroup_categories table migration and model.  It is destined for the master branch, and *should* be the last 
change on the production database.

r? @ncloudioj


ncloudioj:
Looks good to me. r+

Let's also add a `title_bg_color` column to the `tiles` table. Like `bg-color`, it's a VARCHAR(16) with `nullable=False`. After this change, I think we're good to freeze the database changes for the master branch.

tspurway:
I just broke the tests.  Fixing.....

tspurway:
Also, Nan, shouldn't nullable=True, to maintain backward compatibility?  What does firefox do if it gets a payload without titleBgColor?

oyiptong:
It should work just fine without

tspurway:
Yeah, there doesn't seem to be a compelling reason to make it mandatory on the ingest format.

ncloudioj:
@tspurway Yes, you're right. It's optional.

oyiptong:
works just fine without it: https://hg.mozilla.org/integration/fx-team/rev/9d35eca4bc3a#l1.17

tspurway:
let's try another r? @ncloudioj

tests/test_ingest.py
```diff
@@ -501,6 +501,30 @@ def test_adgroups_channel_id_uniqueness(self):
         c = self.env.db.session.query(Adgroup).count()
         assert_equal(32, c)
 
+    def test_title_bg_color(self):
+        """
+        A simple test of frequency caps
```
ncloudioj: A simple test of title_bg_color

tests/base.py
```diff
@@ -4,7 +4,7 @@
 from splice.webapp import create_webapp
 from flask.ext.testing import TestCase
 
-db_uri = os.environ.get('TEST_DB_URI') or 'postgres://localhost/splice_test'
+db_uri = os.environ.get('TEST_DB_URI') or 'postgres://tspurway:tacos@127.0.0.1/splice_test'
```
ncloudioj: Oops...

ncloudioj:
r+ @tspurway

PR #106, opened 2015-09-08
Add category information for adgroups

ncloudioj:


splice/queries.py
```diff
@@ -69,17 +77,18 @@ def tile_exists(session, target_url, bg_color, title, typ, image_uri, enhanced_i
 
     if results:
         for tile_id, adgroup_id in results:
-            # now check frecent sites for this tile
+            # now check frecent sites and adgroup categories for this tile
             db_frecents = get_frecent_sites_for_tile(session, tile_id)
-            if db_frecents == sorted(set(frecent_sites)):
+            db_categories = get_categories_for_adgroup(session, adgroup_id)
+            if db_frecents == sorted(set(frecent_sites)) and db_categories == sorted(set(adgroup_categories)):
```
rlr: I guess there is no way to express these conditions with SQLAlchemy in the giant query above?

ncloudioj: That's a good point. But unfortunately both frencent_sites and adgroup_categories has one-to-many relationship with the adgroup_id. So we have to load them up, then sort and compare in the memory.

rlr:
Looks good to me. Tests pass. r+

ncloudioj:
Thanks @rlr

PR #125, opened 2015-09-23
Latest sp2 Front End

tkiethanom:
r? @r1cky @tspurway

Bootstrap is now an npm module. Initial View Pages from Account down to Tile.

Don't forget to run `npm install` since packages have been added/removed.

The mock API JSON files haven't been kept up to date, still using splice-2.0 API.


front_end/src/actions/AdGroups/AdGroupActions.js
```diff
@@ -0,0 +1,118 @@
...
+		 }).then(response => response.json())
+		 .then((json) => {
+		 dispatch(receiveAddAccount({
+		 'created_at': '',
+		 'email': 'test@gmail.com',
+		 'id': 99,
+		 'name': data.text,
+		 'phone': '+1(888)0000000'
+		 }));
+		 });*/
```
rlr: I guess this ^^ still needs to be fixed later?

rlr:
One thing I just noticed is that all this is using tabs. We like (2) spaces. We probably want to reformat all that at some point.

front_end/src/actions/AdGroups/AdGroupActions.js
tkiethanom: Yea the plan right now is to just make the GETs work. The POSTs and PUTs will come after.

tkiethanom:
Okay I reformatted the code to use 2 spaces and no tab characters.

rlr:
Awesome. I did one pass and things looked pretty good. I'll do another quick pass and merge since this is still a work in progress.

front_end/src/reducers/TileReducer.js
```diff
@@ -0,0 +1,49 @@
...
+  RECEIVE_ADD_TILE,
+  REQUEST_TILES,
+  RECEIVE_TILES,
+  REQUEST_TILE,
+  RECEIVE_TILE
+} from 'actions/Tiles/TileActions';
+
+const initialState = {
+  rows: [],
+  details: [],
```
rlr: I see this pattern in all the models. `rows` is for the list view of all of the instances of the model. `details` is for when we have one instance selected. Is that correct? Although now I am realizing details is an array so I am not sure.

tkiethanom: Correct, `rows` will hold a list of models, and `details` will be a singular model.

tkiethanom: I updated so it should read `details: {}`. I think you're supposed to avoid using null as an initial state anyway since that is the point.

rlr:
OK. Merging this. Thank you!

Issue #121, opened 2015-09-18
Model Changes for Campaign DB


tspurway:
Current Model | New Model | Reason
------------- | --------- | ------
campaigns.locale | adgroups.locale | current campaigns (like booking.com) are run across multiple locales
campaigns.paused | campaigns.paused, tiles.paused, adgroups.paused | parent's paused status overrides childrens'

Deprecated fields: adgroups.[locale, start_date, end_date, start_date_dt, end_date_dt, channel_id], tiles.locale

tspurway:
This issue is also tracking getting the existing model working in sp2

ncloudioj:
The summary of the model change:
 
* New fields: adgroups.[paused, type("directory" | "suggested")], tiles.[paused, status("approved" | "disapproved" | "unapproved")]
* Deprecated fields: adgroups.[start_date, end_date, start_date_dt, end_state_dt, channel_id],
campaings.locale, tile.locale

@tspurway make sense?

tspurway:
this is OK, go ahead

PR #126, opened 2015-09-25
Issue 121 api migration

ncloudioj:
This PR refers to #121, the API migration part. Some highlights of changes:

* remove CRSF
* migrate all APIs based on the new models
* add support for CampaignCountry
* add support for AdgroupCategory
* increase test coverage


splice/models.py
```diff
@@ -95,16 +97,18 @@ class Adgroup(db.Model):
...
     locale = db.Column(db.String(14), nullable=False)
 
-    type = db.Column(db.String(16))
-    paused = db.Column('paused', db.Boolean(), nullable=False, server_default=db.text(u'false'))
+    type = db.Column(db.String(16), nullable=False)
+    paused = db.Column('paused', db.Boolean(), nullable=False, default=False)
     frequency_cap_daily = db.Column(db.Integer())
     frequency_cap_total = db.Column(db.Integer())
-    name = db.Column(db.String(255))
+    name = db.Column(db.String(255), nullable=False)
```
tspurway: We currently have a lot of data in the db that has Adgroup.name == NULL - so this is going to break migrations.  Is it absolutely necessary to have this?

ncloudioj: Yeah, good catch. Will rethink about this. Thought we need a composite unique key here to prevent users from creating duplicate adgroups from a campaign.

Note that for suggested tiles, it became a required field recently as FF needs it in order to get rid of the hardcoded categories.

splice/models.py
```diff
@@ -95,16 +97,18 @@ class Adgroup(db.Model):
     id = db.Column(db.Integer(), autoincrement=True, primary_key=True, info={"identity": [1, 1]})
     locale = db.Column(db.String(14), nullable=False)
 
-    type = db.Column(db.String(16))
-    paused = db.Column('paused', db.Boolean(), nullable=False, server_default=db.text(u'false'))
```
tspurway: We have existing data where Adgroups.type == NULL.  If this is necessary we need a strategy

splice/models.py
```diff
@@ -23,11 +22,12 @@ class Campaign(db.Model):
...
     start_date = db.Column('start_date', db.DateTime(), nullable=True)
     end_date = db.Column('end_date', db.DateTime(), nullable=True)
-    name = db.Column('name', db.String(length=255), nullable=True)
-    paused = db.Column('paused', db.Boolean(), nullable=False, server_default=db.text(u'false'))
+    name = db.Column('name', db.String(length=255), nullable=False)
+    paused = db.Column('paused', db.Boolean(), nullable=False, default=False)
     channel_id = db.Column('channel_id', db.Integer(), db.ForeignKey("channels.id"))
     account_id = db.Column('account_id', db.Integer(), db.ForeignKey("accounts.id"))
     created_at = db.Column('created_at', db.DateTime(), server_default=db.func.now(), nullable=False)
+    __table_args__ = (db.UniqueConstraint('account_id', 'name'),)
```
tspurway: This unique constraint will also prevent migrations from working on our existing data.

splice/models.py
```diff
@@ -85,6 +85,8 @@ class Tile(db.Model):
     status = db.Column(db.String(16), nullable=False, server_default=u'unapproved')
 
     created_at = db.Column(db.DateTime(), nullable=False, server_default=db.func.now())
+    __table_args__ = (db.UniqueConstraint('adgroup_id', 'title', 'target_url', 'image_uri',
+                                          'enhanced_image_uri', 'type'),)
```
tspurway: This unique constraint also will cause migrations to fail on our production data

splice/models.py
ncloudioj: I agree. I thinks it's also necessary to name every single adgroup, even for those DT ones.

We can fill up the existing data with the default value in the migration script.

splice/queries/campaign.py
```diff
@@ -0,0 +1,74 @@
...
+
+def get_campaigns(account_id=None):
+    from splice.environment import Environment
+
+    env = Environment.instance()
+
+    query = (
+        env.db.session
+        .query(Campaign, CampaignCountry.country_code)
+        .join(CampaignCountry, Campaign.id == CampaignCountry.campaign_id)
```
tspurway: This join is going to produce multiple rows per campaign, which I don't think is what you want.  Maybe do not return countries in this query, but add another path /campaigns/25/countries

PR #137, opened 2015-10-01
Tweak the model change to keep backward compatibility

ncloudioj:
r? @tspurway


migrations/versions/3cc07fb28766_.py
```diff
@@ -8,7 +8,7 @@
 
 # revision identifiers, used by Alembic.
 revision = '3cc07fb28766'
-down_revision = '16fa35dd63a2'
+down_revision = None
```
tspurway: isn't there a down migration?

ncloudioj: Oops, that was a mis-kill. It's now fixed.

tspurway:
r+

the migration script works on this, thanks, Nan!

ncloudioj:
Awesome! Merging it now.

PR #141, opened 2015-10-02
Use Flask-Cors to support CORS

ncloudioj:
Add Flask-Cors library to support CORS, the built-in support from Flask-Restful can't handle the failure cases properly.


rlr:
I like better! r+!

Issue #142, opened 2015-10-02
Using ISO8601 datetime in all API


ncloudioj:
When server and client exchanging data with each other, we'll be using ISO8601 format for datetime in UTC, i.e '2015-10-02T01:38:11+00:00'. or '2015-10-02T01:38:11.322406' omit the TZ info.

PR #145, opened 2015-10-05
Unify time format to ISO8601

ncloudioj:
This PR intends to unify the time format for all SP2 models. See #142 .

**NOTE:** you might need to set the `timezone` for your Postgresql to 'UTC'. Otherwise, it'll automatically do the conversion regardless the timezone you've passed to it.
 
In your `postgresql.conf`, search `timezone` and replace it with 'UTC'. It's already configured properly in production.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,3 +18,4 @@ bleach==1.4.1
 tld==0.7.3
 flask-restful==0.3.4
 Flask-Cors==2.1.0
+pytz==2015.6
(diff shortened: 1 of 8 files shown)


rlr:
Do we care about the coverage check? They seem to be stuck in pending recently

rlr:
Otherwise, this looks good and tests pass. r+

tspurway:
yep r+

ncloudioj:
Yes, the coverage seems unhappy probably because of the front-end tests. I'll take a look at this as we do want to keep the coverage for the back-end part.

Issue #144, opened 2015-10-05
Additional APIs needed for Campaigns


tkiethanom:
I need to be able to grab all Channels and Countries to fill dropdowns for creating and updating Campaigns.

ncloudioj:
Yup, will add those ones shortly.

tspurway:
You will also need a list of locales.  Let's pay attention to the ordering of the country and locales - they should be ordered by popularity (US, CA, UK, ...) (en-US, es-US, fr-FR, etc.) not alphabetically

tkiethanom:
In the current API, locale is a field of Ad Groups.

tspurway:
@tkiethanom - good point - we won't need locales

PR #148, opened 2015-10-06
Add support APIs for channel, country, and locale

ncloudioj:
This PR adds an API endpoint to support the campaign&adgroup management. Specifically, it exposes `/api/init/<string:target>`, in which `target` could be one of,
* "countries", all the country codes and country names
* "locales", all the locales served in Firefox
* "channels", all the channels and the related channel_ids
* "all", get all of targets above

**Note** that all of those targets are read-only, therefore, only HTTP GET method is available for now.

### DIFF
--- a/migrations/versions/16fa35dd63a2_.py
+++ b/migrations/versions/16fa35dd63a2_.py
@@ -14,6 +14,7 @@ depends_on = None
 
 from alembic import op
 import sqlalchemy as sa
+from splice.environment import Environment
 
 
 def upgrade(engine_name):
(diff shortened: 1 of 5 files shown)


ncloudioj:
@tkiethanom FYI, this refers to #144

rlr:
This looks good to me. Tests pass. r+ as long as it meets @tkiethanom's needs.

tkiethanom:
Is there an endpoint for getting the name of a Channel by id?

tkiethanom:
Or should I be getting all the Channels and looping through it on my end?

ncloudioj:
@tkiethanom Yes, all the targets from this endpoint could be cached on the client side. For example, there are currently four items in the Channels table. There is no need to fetch that from backend regularly.

tkiethanom:
Okay that works for me!

PR #149, opened 2015-10-07
Disable front-end test

ncloudioj:
Investigating issues for Coveralls. Please ignore this PR.


ncloudioj:
This gets fixed now. The reason was after running tests in front_end, it should return back to top-level directory so that TravisCI could send coverage data to Coveralls.

Issue #150, opened 2015-10-07
Add API for fetching categories for adgroups and tiles


ncloudioj:
The client will need this while creating suggested adgroups&tiles.

Note that the target sites for each category won't be sent by the server, instead, the server will populate the `frecent_sites` field based on the `category` for that adgroup.

The endpoint for this: "api/init/categories", it will be bundled in the "api/init/all" as well.

PR #152, opened 2015-10-09
Add API for fetching categories

ncloudioj:
This refers to #150.

Note that we've added a fixture from `Bucketer` for the mapping between categories and frecent sits. This might come from another source in the future.

A new field "adgroup_name" has been added to the fixture as well. It's a required field for all the suggested tiles, which allows Firefox to eliminate the hard-coded URLs and adgroup names.


rlr:
looks good to me. tests are happy. r+!

Issue #113, opened 2015-09-15
Port existing Splice UI to new stack


tspurway:
This feature is to refactor our existing AngularJS UI in Splice to use our new stack - feature-for-feature.

ncloudioj:
According to the new biz requirement, we're looking to add country as one of targets for content suggested tiles. E.g. running content tiles for locale "en-US" in "US". That will significantly decrease the size of tile set (we used to run content tiles for all "en-US" FF users), thus, the slow tile loading as well as the massive memory usage will no longer be bugging us in the near future.

Depends on where do we stand in the development, I think that we can put the UI enhancement on hold, and focus on the "Approval Workflow" and the related UI now. What do you guys think? @tspurway @rlr

rlr:
I am game either way. I'll assume we will continue down this path unless there is something else that takes priority?

tspurway:
I think we continue on.  There are a number of advantages to moving our codebase to a single UI library and make some incremental improvements.  We have some time now, let's continue on.

ncloudioj:
Sounds good. Let's wrap it up then.

rlr:
Filed pull request #153

ncloudioj:
awesome. i am on it.

PR #153, opened 2015-10-14
Redux frontend for existing splice UI (Issue #113)

rlr:
Lots of commits and files. Oh my... I built the entire thing in a new directory `ui/`. I'm game to leave it there, rename it, move it into `splice/` to replace the angular stuff once it's good to go or sounds good.

To give it a spin:

```
$ cd ui
$ npm install
$ npm start
```

In another shell: `$ manage.py runserver`

Point your browser at: `localhost:3000`

I need to figure out how to write some meaningful tests next. I'll probably follow what TBG is doing.

r? @ncloudioj @tspurway


ui/src/actions/Authoring.js
```diff
@@ -0,0 +1,307 @@
...
+
+   var url = state.distribution.publishUrl;
+   url = url + '?' + encodeQueryParams(params);
+
+   fetch(url, {
+     method: 'POST',
+     headers: {
+       'Accept': 'application/json',
+       'Content-Type': 'application/json'
+       //,'X-CSRFToken': document.querySelector('meta[name=csrf-token]').attributes['content'].value
```
ncloudioj: Let's remove this line as CSRF is already disabled.

ui/src/components/Tiles.js
```diff
@@ -0,0 +1,69 @@
...
+    var rows = [];
+    var tilesType = this.props.tilesType;
+    this.props.tiles.forEach(function(webtile) {
+      var thumbStyle = {
+        backgroundImage: 'url(' + webtile.imageURI + ')'
+      };
+      var enhancedThumbStyle = {
+        backgroundImage: 'url(' + webtile.enhancedImageURI + ')'
+      };
+      var frecent = [];
```
ncloudioj: Rather than displaying the lengthy `frecent_sites` of the suggested tiles here, listing a few other fields would be more useful for the traffickers/viewers. Such as,

"adgroup_categories": ["Technology General", "Technology Mobile", ...],
"frequency_caps": {"daily": 1, "total": 5},
"explanation": "A technology content site suggestion"

ui/src/components/Tiles.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+Tiles.propTypes = {
+  tiles: PropTypes.arrayOf(
+    PropTypes.shape({
+      title: PropTypes.string.isRequired,
+      imageURI: PropTypes.string.isRequired,
+      enhancedImageURI: PropTypes.string,
+      type: PropTypes.string,
```
ncloudioj: "type" is also a required field for both directory and suggested tiles.

ui/src/reducers/Authoring.js
```diff
@@ -0,0 +1,151 @@
...
+      return Object.assign({}, state, {
+        selectedType: action.tileType
+      });
+    case AUTHORING_SET_PUBLISH_DATE:
+      return Object.assign({}, state, {
+        scheduled: action.moment
+      });
+    case AUTHORING_SET_DEPLOY_NOW:
+      return Object.assign({}, state, {
+        deployNow: action.deployNow
```
ncloudioj: While testing the "DeployNow" feature, the argument string seems incorrect. Therefore, the UI showed the deploy status as "No" afterwards. 

The expected url should be "/api/authoring/all_tiles?deploy=1&channelId=1", but the actual one was "/api/authoring/all_tiles?deploy=true&channelId=1". "deploy=true" made the server ignore the "deploy now" command.

Perhaps this action has passed a wrong argument "deployNow" here.

rlr: Nice catch. fixed: 306c400361df6e25becc14af869a4729cf446051

ui/src/actions/Authoring.js
rlr: :+1: b45e028cda5eb0c9d3c8d4b046bc650337f4fef2

ncloudioj:
That's all I found so far. All the business looks good to me in the Authoring/Upcoming actions and reducers :+1: 

Another feature we can add later on is that being able to view the previous distributions. I think it's quite similar as what we have in the upcoming page, except we're looking back.

ui/src/components/Tiles.js
rlr: Made type required: 7b30fa7742c16b9264e66f20ae3c8902c7fb1305
Added adgroup_categories and frequency_caps: ff4c8fc9ba77f6b1146247dc028760d4c00e4b85

ui/src/components/Tiles.js
rlr: Now showing categories and caps too: fd0fa962de073a900e9f242f7b46cdcdc4b5df6d

PR #170, opened 2015-10-29
Add support for the bulk upload

ncloudioj:
This PR intends to implement the bulk upload feature for Splice 2.0.

In a nutshell, given the inputs from splice user, it ingests the data, performs the image processing, then uploads the new images to S3 and insert new adgroups and tiles to database.

The input of this API (api/campaigns/<campaign_id>) includes a image zip file, a tsv meta file, and the campaign id. Two sample files have been attached in the `splice/tests/fixtures/` for the testing purpose.


ncloudioj:
Just ignore the conflicts for now. Will resolve them later.

r? @tspurway or @rlr

ncloudioj:
A few more comments for this PR:
* this feature is mainly designed for the Suggested Tiles (ST), to support Directory Tiles, we'll need to pass more arguments
* a better auto generated adgroup name. In this implementation, it's using "Adgroup {category} {3-digit random code}", e.g. "Adgroup Technology_New 018". Perhaps using the next adgroup id from the database is a better idea here, like "Adgroup 118"
* this API is **not** async, it'll hang the request until it finishes or terminates from errors. We've discussed about this concern before, it should be fine for now

requirements.txt
```diff
@@ -19,3 +19,4 @@ tld==0.7.3
 flask-restful==0.3.4
 Flask-Cors==2.1.0
 pytz==2015.6
+Pillow==3.0.0
```
rlr: makes me sleepy

ncloudioj: LOL, I've been sleepy since the day 1 in NOL.

splice/web/api/campaign.py
```diff
@@ -126,5 +131,40 @@ def put(self, campaign_id):
...
+def handler_bulk_upload(campaign_id):
+    """bulk uploading adgroups and tiles with the given campaign_id."""
+
+    def is_allowed_files(name, allowed_sets):
+        return '.' in name and name.rsplit('.', 1)[1] in allowed_sets
+
+    creatives = request.files["creatives"]
+    assets = request.files["assets"]
+    if not is_allowed_files(creatives.filename.lower(), _VALID_CREATIVE_EXTENSIONS) \
+            or not is_allowed_files(assets.filename.lower(), _VALID_ASSETS_EXTENSIONS):
```
rlr: why don't you do the `.filename.lower()` inside `is_allowed_files()`? So you pass the file to `is_allowed_files()` and it checks the extension. That would clean this up a little. 

And I think it should be `is_allowed_file()` without the `s` since you only check one file at a time.

splice/web/api/campaign.py
```diff
@@ -126,5 +131,40 @@ def put(self, campaign_id):
...
+            return jsonify(message="Invalid files uploaded."), 400
+
+    campaign = get_campaign(campaign_id)
+    if campaign is None:
+        return jsonify(message='Campaign not found.'), 404
+
+    (_, tmp_creatives) = tempfile.mkstemp()
+    (_, tmp_assets) = tempfile.mkstemp()
```
rlr: Those parens look weird. Pretty sure you don't need that.

splice/web/api/campaign.py
```diff
@@ -126,5 +131,40 @@ def put(self, campaign_id):
...
+    creatives.save(tmp_creatives)
+    assets.save(tmp_assets)
+    try:
+        bulk_upload(tmp_creatives, tmp_assets, campaign_id, campaign["channel_id"])
+    except Exception as e:
```
rlr: you need to delete the temp files in this case too.

splice/web/api/campaign.py
```diff
@@ -126,5 +131,40 @@ def put(self, campaign_id):
...
+            return jsonify(message="Invalid files uploaded."), 400
+
+    campaign = get_campaign(campaign_id)
+    if campaign is None:
+        return jsonify(message='Campaign not found.'), 404
+
+    (_, tmp_creatives) = tempfile.mkstemp()
+    (_, tmp_assets) = tempfile.mkstemp()
+    creatives.save(tmp_creatives)
+    assets.save(tmp_assets)
```
rlr: Actually, you are saving these files to disk and reopening them again down in load assets. Why not just pass the files down. `zipfile` can take a file object instead of a path. And you can avoid doing the `open(assets, "rb") as f`. Would probably make everything a bit faster.

splice/web/api/campaign.py
```diff
@@ -126,5 +131,40 @@ def put(self, campaign_id):
...
+    try:
+        bulk_upload(tmp_creatives, tmp_assets, campaign_id, campaign["channel_id"])
+    except Exception as e:
+        return jsonify(message="Error: %s" % e), 400
+    else:
```
tspurway: should this be `finally` instead of else?

ncloudioj:
Oh, boy. So many bugs you guys have spotted. Thank god, it's not a rocket launcher app :)

Repairing now...

rlr:
haha. not bugs. just nits and potential improvements :wink:

splice/web/api/campaign.py
ncloudioj: Yup, you're totally right. I've tried that but failed :(. Storing everything to the tmp dir bought me more time on the trickier ingestion/S3/database part. Let's turn back this now. Thanks amigo, good catch :+1:

tspurway:
r+

ncloudioj:
Closed in 0e04193

PR #172, opened 2015-10-30
Sp2 Front End Campaign Filters

tkiethanom:
r? @rlr @tspurway 

Added UI to handle latest filters (past, scheduled and in_flight) added to Campaigns.


front_end/src/components/Campaigns/CampaignList/CampaignRow.js
```diff
@@ -1,28 +1,52 @@
...
+    countries.map((row, index) => {
+      if (index > 0) {
+        output += ', ';
+      }
+
+      country = getCountry(row, this.props.init_countries);
+      if (country !== undefined && country.country_name !== undefined) {
+        output += country.country_name;
+      }
+    });
```
rlr: This is a little messy and it looks like there are cases where you could end up with `country1, , country2`. I would build an array of country names then:

```
>> ['x', 'y', 'z'].join(', ')
"x, y, z"
```

rlr:
Other than that, this looks good to me.

tkiethanom:
Updated this to use array.join()

tkiethanom:
Commented out Approvals fixes #176

rlr:
:+1: r+

PR #186, opened 2015-10-30
Make Clang happy on Mac OS

ncloudioj:
r? @oyiptong

### DIFF
--- a/setup-project.sh
+++ b/setup-project.sh
@@ -9,6 +9,7 @@ then
     set -x
     echo "setting up development virtualenv"
     export MOZ_SPLICE_DEV=1
+    export CFLAGS='-std=c99'
 else
     echo "setting up production virtualenv"
 fi


oyiptong:
Can you just set the CFLAGS environment variable at the beginning of the file?

oyiptong:
i.e. 
```
export CFLAGS='-std=c99'
```
actually, it might be safer to set that variable only for development and not for production, i.e. not at the top of the file.

oyiptong:
we probably use gcc in our jenkins build. i wonder if that flag will have any adverse effects

ncloudioj:
Yes, I don't think we need this for the Linux box. So I just dropped an alias only for development.

oyiptong:
how about an environment variable in development mode?

BTW, i just checked, gcc does support the `-std=c99` CFLAGS parameter

ncloudioj:
No problem, new change pushed.

oyiptong:
r+

Issue #174, opened 2015-10-30
Landing page for Splice 2


tspurway:
We currently have two distinct UIs in Splice: the distribution UI (`./ui`) and the Campaign Mgmt UI (`./front_end`).  We need a deployment strategy so that users can easily navigate between these UIs.

This could be a landing page, or a navigation element added to all the existing pages.  Ideas?

rlr:
Right now these are two separate apps, but I am guessing longterm we'll only have one app. We would merge them, tossing whatever we don't need anymore from the old UI.

That said. For now... I guess we would need to have them live at two separate paths? Something like this maybe:

* `/` has the landing page
* `/old/` has the app in `./ui`
* `/sp2/` has the app in `./frontend`

Thoughts?

tkiethanom:
If you want to direct users to a single url a landing page would be useful. It's also probably likely users will want to jump back and forth between them. So both.

rlr:
If we expect a lot of back and forth, then maybe I should go ahead and move existing UI into the sp2 app?

rlr:
landed be7fb967c9e1dfa060d3ae8657283cb9fd5ac17d

Issue #188, opened 2015-11-02
Bring back ingest, authoring and upcoming api to the sp2 branch


rlr:
I just realized that we nuked ingest.py, authoring.py, upcoming.py and their respective tests. We weren't planning on using the existing UI when we started down this route. We need to bring them back.

Are those the only files that we need to bring back? @ncloudioj

rlr:
The commit those were removed is here: 7fa60ccc5027d3282d9f4ce906ecc6c85f34ddd0

Unfortunately there are other changes that we need to keep so it's not a simple revert.

Issue #173, opened 2015-10-30
Building JS assets bundle in Fabric


tspurway:
Our deployment process includes a packaging step using `fab package` which ends up running the Fabric file `./fabfile.py`.  

During the package phase, we need to build JS bundles for both `./ui` and `./front_end`.

@k88hudson and @rlr: what is a good strategy for integrating WebPack with Fabric?

rlr:
We would have fabric do something like:

```
$ cd ui
$ npm install
$ npm run-script bundle
```

And then it can move the bundle files around if needed or we can configure webpack to put them in the right place from the getgo.

k88hudson:
If you wanted to, you could add `npm run bundle` as a `postinstall` script as well.

Also @rlr you don't need to call `./node_modules/webpack/bin/webpack.js` in your script, npm adds webpack's bin directory to the path automatically -- you can just call `webpack` in your script

k88hudson:
i agree that calling npm from fabric is the best approach :+1:

rlr:
There is now a `build_ui` command in the fabric file. 56d40fa802c5a939fb0ce99bb9f628b284b3849f

PR #189, opened 2015-11-02
Fix Issue #174: Splice Landing Page.

rlr:
Just some basic unstyled HTML for now.

The distributions app doesn't work because the api is missing in this branch (Issue #188).

The new splice 2 app just has a placeholder template. I need to work on it's bundling/deploying so it ends up in the right spot (Issue #173).

r? @ncloudioj @tspurway


rlr:
this is no longer valid since sp2 branch doesn't have the new ui in ./ui anymore. Closing.

PR #196, opened 2015-11-04
Fix #195: Add db side defaults for created_at where missing.

### LINKED ISSUE #195: In the Campaign Database, incorrect DDL for `created_at` column
The column should never be set by any client code, it should always be set using a default timestamp from the server.  This means we should also disable access to this column from all APIs

rlr:
r? @ncloudioj


ncloudioj:
r+ :+1:

PR #211, opened 2015-11-10
Landing page and fabric script to build redux apps

rlr:
Fixes issue #173 and issue #174.

To try it out (npm and gulp are required):

    $ fab build_ui
    $ manage.py runserver
    $ open http://localhost:5000/

For the new splice ui (campaign manager), there is a problem with background images. They are specified in SASS as absolute paths like `/public/img/filename.jpg` and we are serving them from `/static/build/....`. That's an outstanding issue I'll need to figure out with @tkiethanom. But this gets us closer.

r? @ncloudioj @tspurway


rlr:
NOTE: all the code in ui/ had been previously reviewed in the master branch.

fabfile.py
```diff
@@ -92,6 +95,18 @@ def deploy_cold():
     set_symlinks()
 
 
+def build_ui():
+    with lcd('ui'):
+        local('pwd')
+        local('npm install')
+        local('npm run-script bundle')
```
k88hudson: you can use `run` instead of `run-script`

fabfile.py
```diff
@@ -92,6 +95,18 @@ def deploy_cold():
...
+
+    with lcd('front_end'):
+        local('npm install')
+        local('gulp build:dist')
```
k88hudson: This requires a global install of `gulp`. Wouldn't it be better if we just installed it locally and called this command in the script?

fabfile.py
rlr: oops. just realized i left in a PWD

fabfile.py
rlr: That's how TBG has set up the new app. We can change it later. I'd like to remove the gulp dependency.

k88hudson: I don't mean switch from gulp completely, just this: https://github.com/oyiptong/splice/pull/215/files

rlr:
Fixed those comments ^^. They depend on landing pull #215.

rlr:
I've rebased again. Should be good for another r? pass now.

ncloudioj:
It looks good to me. r+!

We can drop some "# pragma: no cover" in the `view.py` to make the coveralls happy.

PR #213, opened 2015-11-11
Use the dev bucket by default

ncloudioj:
In Sp2, we should always use the dev bucket on S3 by default. Nobody should touch the bucket for production in any case.

This is safer than letting developers overwrite the bucket in their `localconfig.py`. Also, we always explicitly set the bucket name in production, the dev bucket would be overwritten at that point.


tspurway:
r+

PR #216, opened 2015-11-11
Ingest from the database directly

ncloudioj:
This PR intends to replace the JSON based tile generation. It fetches all the tiles from the database based on the campaigns' flight date. Then it groups tiles by channels, country, and locales, generating the index files accordingly. The file format is backward compatible with the current Version 3 one. The only difference is it'll contain all the tiles instead of only one sponsored tile in the current distribution.

Two API endpoints are exposed here:
HTTP GET   /api/distributions\?channel_id=3\&date="2015-10-10"
HTTP POST /api/distributions

Where the GET returns the index files in JSON, and POST pushes all the index files to S3

### DIFF
--- a/splice/queries/distribution.py
+++ b/splice/queries/distribution.py
@@ -1,40 +1,55 @@
-from collections import defaultdict
-from collections import namedtuple
 import hashlib
 import os
 import urllib
-
-from datetime import datetime
 import json
 
+from datetime import datetime
+from collections import defaultdict
+from collections import namedtuple
+
+from sqlalchemy.sql.expression import false
 from splice.models import Tile, Adgroup, Campaign, CampaignCountry
+from splice.web.api.tile_upload import load_bucketer
+
 
 Dists = namedtuple("Dists", ['legacy', 'directory', 'suggested'])
 
 
-def _tile_dict(tile, legacy=False):
-        tile_dict = dict(
-            target_url=tile.target_url,
-            bg_color=tile.bg_color,
-            title_bg_color=tile.title_bg_color,
-            title=tile.title,
-            typ=tile.type,
-            image_uri=tile.image_uri,
-            enhanced_image_uri=tile.enhanced_image_uri,
-            locale=tile.adgroup.locale,
-            channel_id=tile.adgroup.channel_id)
-        if not legacy:
+def _tile_dict(tile, bucketer, legacy=False):
+    tile_dict = dict(
+        directoryId=tile.id,
+        url=tile.target_url,
+        bgColor=tile.bg_color,
+        title=tile.title,
+        type=tile.type,
+        imageURI=switch_to_cdn_url(tile.image_uri),
+        enhancedImageURI=switch_to_cdn_url(tile.enhanced_image_uri))
+    if not legacy:
+        # only suggested tiles have categories
+        if tile.adgroup.categories:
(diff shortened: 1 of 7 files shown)


ncloudioj:
**NOTE**: don't try the HTTP POST endpoint until we merge the #213, in case it accidentally overwrites the index file in production.

r? @tspurway and @rlr

splice/queries/distribution.py
```diff
@@ -0,0 +1,134 @@
...
+        type=tile.type,
+        imageURI=switch_to_cdn_url(tile.image_uri),
+        enhancedImageURI=switch_to_cdn_url(tile.enhanced_image_uri))
+    if not legacy:
+        # only suggested tiles have categories
+        if tile.adgroup.categories:
+            # TODO(najiang@mozilla.com) only grab the first category here.
+            # We should create multiple tiles for multiple categories, with
+            # each tile has a frecent site list based on the category.
+            bucket = bucketer[tile.adgroup.categories[0].category]
```
rlr: I read the comment above but still don't understand why you only use the first category.?

splice/queries/distribution.py
```diff
@@ -0,0 +1,134 @@
...
+                frequency_caps=frequency_caps,
+                adgroup_name=bucket["adgroup_name"],
+                adgroup_categories=[category.category for category in tile.adgroup.categories],
+                explanation=tile.adgroup.explanation,
+                check_inadjacency=tile.adgroup.check_inadjacency))
+    return tile_dict
+
+
+def switch_to_cdn_url(image_uri):
+    """See https://github.com/oyiptong/splice/issues/203"""
```
rlr: that should probably be a comment instead of a docstring.

splice/queries/distribution.py
```diff
@@ -0,0 +1,134 @@
...
+    if today is None:
+        today = datetime.utcnow().date()
+
+    query = (
+        env.db.session
+        .query(Tile)
+        .filter(Tile.paused == false())
+        .filter(Adgroup.paused == false())
+        .filter(Campaign.paused == false())
+        .filter(Campaign.end_date > today)
```
rlr: I guess end date is not inclusive. If it starts today and ends tomorrow it will only run today. That's fine as long as it's understood everywhere (UI).

splice/web/api/distribution.py
```diff
@@ -0,0 +1,52 @@
...
+
+class DistributionAPI(Resource):
+    def __init__(self):
+        super(DistributionAPI, self).__init__()
+
+    def get(self):
+        """Returns the distributions on a specific date"""
+        args = arg_parser.parse_args()
+        artifacts = get_possible_distributions(today=args.get('date'),
+                                               channel_id=args.get('channel_id'))
```
rlr: this is perfecto for the UI I need to build :+1:

rlr:
I did one pass and looks pretty good. Just minor questions and comments.

splice/queries/distribution.py
ncloudioj: Good catch! Intuitively, most people prefer to use the inclusive style flight dates.

ncloudioj:
Thanks @rlr. Pushed a new commit to clean up the comments, fill up the TODO hole.

rlr:
looks good to me. r+ from me

Issue #108, opened 2015-09-10
Support multiple concurrent directory tiles (scheduling)


tspurway:
[Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1193822)

The implementation of this is entirely server-side, and will make changes to both Splice and Onyx.

Splice 
======

When splice creates the distribution files, it will need a way of supporting multiple distributions for a single GEO/LOCALE.  In addition, we will need to store the locations of these multiple tile sets in the 'index' file that is persisted into S3.

Onyx
====

Onyx will need to read the new index format that has multiple possible distributions, and to choose one *at random* when a Firefox user in a matching GEO/LOCALE performs a daily 'fetch' operation for their tiles.

oyiptong:
Unsure if this is a worthwhile feature. We can get this implemented client-side once remote-newtab pages land.

The cost of this implementation needs to be estimated; it will have a short shelf-life.

oyiptong:
How much overlap is there to do both client-side and server-side decisioning? In theory, the ingestion aspects are the same.

Client-Side Decisioning
==================
In the client-side decisioning scenario, we'll want a new API endpoint with minimal changes to the way onyx currently work. It would merely send a new payload format with multiple possible choices for sponsored tiles.

Server-Side Decisioning
==================
In the server-side scenario, onyx will need to do more work.

There are actually two scenarios:
1. multiple distributions are generated by splice, onyx just makes a choice at random and serves
2. only one distribution is generated. For versions of Firefox that can't do client-side decisioning, it would need to generate a distribution on the fly.

Given that this is a temporary fix, 2 seems like an acceptable computational cost. That minimizes work on the splice-side of things going forward since multiple-distribution generation will be obsolete quite quickly.

Since we have ASG, the increased computational cost from JSON serialization/deserialization won't have much operational impact. Plus it's only a temporary cost.

ncloudioj:
I agree with @oyiptong that we'd better implement this feature through the upcoming RNTP. It only needs a minimal change to the current back-end applications.

One major downside of server-side decisioning is that it'll be rather annoying to implement and deploy features like weighted randomized serving or any other non-trival strategies. That is going to depend on two moving parts (Splice&Onyx).

I'd recommend that let Splice ship all the sponsored tiles to FF, and let the client decide which one will be served each time, based on the logic inside RNTP.

tspurway:
So we can use both server-side and client side decisioning.  

[...]

PR #236, opened 2015-11-17
Issue 108 multiplex directory tiles

ncloudioj:
This PR implements the feature for rotating directory tiles. See more details in #108 

You can use following API endpoints in your dev environment. Note that both `channel_id` and `date` are optional arguments. 

```shell
$ http GET :5000/api/distributions\?channel_id=1\&date="2015-10-04"
$ http POST :5000/api/distributions\?channel_id=1\&date="2015-10-04"
```

### DIFF
--- a/splice/queries/distribution.py
+++ b/splice/queries/distribution.py
@@ -173,11 +173,14 @@ def multiplex_directory_tiles(tiles):
     """Directory tile multiplexer that creates all possible combinations of the
     tile sets based on its type and the target url. Given multiple sponsored tiles,
     it'll pick one of them and create a new tile set with other directory tiles.
-    It also multiplexes the tiles with the same TLD+1 target url (e.g. www.mozilla.org).
-    Therefore, given a tile set with N sponsored tiles and M identical TLD+1 tiles, the
+    It also multiplexes the tiles with the same Full Qualified Domain Name(FQDN).
+    For example, "https://www.mozilla.org/pocket" and "https://www.mozilla.org/gears"
+    share the same FQDN "www.mozilla.org".
+
+    Therefore, given a tile set with N sponsored tiles and M identical FQDN tiles, the
     total number of multiplexed directory tile set is N*M.
 
-    Note: this function will NOT handle the same tld+1 URL between the sponsored and
+    Note: this function will NOT handle the same FQDN between the sponsored and
     non-sponsered tiles.
 
     Params:
@@ -187,21 +190,21 @@ def multiplex_directory_tiles(tiles):
         3rd entry in the list, other tiles will be sorted by the tile id.
     """
     sponsored = []
-    tlds = defaultdict(list)
+    fqdns = defaultdict(list)
     for tile in tiles:
         if tile["type"] == "sponsored":
             sponsored.append(tile)
         else:
             url = furl(tile["url"]).netloc
             assert url, "URL: %s is invalid" % tile["url"]
-            tlds[url].append(tile)
+            fqdns[url].append(tile)
 
     # if no sponsored tiles found, use a sentinel instead
     if not sponsored:
         sponsored.append(_SENTINEL)
 
-    # Cartesian product of the identical tld tiles and the sponsored tiles
-    for multiplex in product(sponsored, *tlds.values()):
+    # Cartesian product of the identical FQDN tiles and the sponsored tiles
+    for multiplex in product(sponsored, *fqdns.values()):
         copy = list(multiplex)
         sponsored_tile, rest = copy[0], copy[1:]
         rest.sort(key=lambda tile: tile["directoryId"])
--- a/splice/queries/distribution.py
+++ b/splice/queries/distribution.py
@@ -125,27 +125,35 @@ def get_possible_distributions(today=None, channel_id=None):
     tile_index = {}
     for (channel, country, locale), (legacy, directory, suggested) in tiles.items():
         country_locale = "%s/%s" % (country, locale)
+        legacy_keys, ag_keys = [], []
+
         # v2
-        legacy_json = json.dumps({locale: legacy}, sort_keys=True)
-        legacy_hsh = hashlib.sha1(legacy_json).hexdigest()
-        legacy_key = "{0}/{1}.{2}.json".format(channel, country_locale, legacy_hsh)
-        artifacts[channel].append({
-            "key": legacy_key,
-            "data": legacy_json})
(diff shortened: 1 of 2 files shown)


splice/queries/distribution.py
```diff
@@ -154,3 +164,48 @@ def get_possible_distributions(today=None, channel_id=None):
...
+
+
+_SENTINEL = object()
+
+
+def multiplex_directory_tiles(tiles):
+    """Directory tile multiplexer that creates all possible combinations of the
+    tile sets based on its type and the target url. Given multiple sponsored tiles,
+    it'll pick one of them and create a new tile set with other directory tiles.
+    It also multiplexes the tiles with the same TLD+1 target url (e.g. www.mozilla.org).
```
rlr: is that really what you want? then it will combine support.mozilla.org, developer.mozilla.org, contribute.mozilla.org, etc all into one? So that's a change from the current setup where by default I am seeing a tile for all of those at once.

rlr: So, for example, this wouldn't be possible anymore:

tests/api/test_distribution.py
```diff
@@ -58,3 +59,39 @@ def test_post_all_distributions_failure(self, query_mock):
...
+        tile4 = {"type": "affiliate", "directoryId": 4, "url": "http://www.mozilla.org"}
+        tile5 = {"type": "affiliate", "directoryId": 5, "url": "http://developer.mozilla.org"}
+        tiles = [tile1, tile2, tile3, tile4, tile5]
+        ret = list(multiplex_directory_tiles(tiles))
+        assert_equal(len(ret), 4)
+        assert_equal([
+            [tile3, tile5, tile1],
+            [tile4, tile5, tile1],
+            [tile3, tile5, tile2],
+            [tile4, tile5, tile2]
```
rlr: oh hmm, maybe you really meant TLD+2 ? The test here looks like it.

splice/queries/distribution.py
ncloudioj: it's "TLD+1", we only want to multiplex those tiles with the same "TLD+1", for example, "www.mozilla.org/pocket", "www.mozilla.org/principles", ..., "www.mozilla.org/gears". Both of them have "TLD+1" as "www.mozilla.org"

splice/queries/distribution.py
```diff
@@ -154,3 +164,48 @@ def get_possible_distributions(today=None, channel_id=None):
...
+    """
+    sponsored = []
+    tlds = defaultdict(list)
+    for tile in tiles:
+        if tile["type"] == "sponsored":
+            sponsored.append(tile)
+        else:
+            url = furl(tile["url"]).netloc
+            assert url, "URL: %s is invalid" % tile["url"]
+            tlds[url].append(tile)
```
rlr: Actually, it looks like you are using the full domain name... So I am not sure what you meant by TLD+1. I thought it meant the TLD (for example, `.com`) + one subdomain level (for example, `mozilla`).

But, you are doing this:

```
In [4]: furl('http://a.bunch.of.subdomains.support.mozilla.org').netloc
Out[4]: 'a.bunch.of.subdomains.support.mozilla.org'
```

rlr:
This looks good. The wording in the docstring is confusing though. Worth fixing that.

tests/api/test_distribution.py
ncloudioj: heck, that was TLD+2? My apologies for the confusion. So I kept using this term wrongly. Anyway, that's how Firefox groups the url histories.

rlr:
I think you made up the term "TLD+X" or at least my google-fu is failing to teach me what it means :stuck_out_tongue_winking_eye:

ncloudioj:
actually, this is where i learnt this TLD thing from, https://pypi.python.org/pypi/tld

rlr:
So, you changed the wording to TLD+2 but that isn't what the code is doing. Why don't you say you are multiplexing by the full domain name (aka, netloc) which is what the code does? Some good examples would help too.

rlr:
Wikipedia suggests this is called FQDN: https://en.wikipedia.org/wiki/Fully_qualified_domain_name

oyiptong:
yeah. examples of TLD (top-level domains) are: com, net, org.
ccTLD's (country-code top-level domains) are: fr, uk, ch, us

What we use for Tiles are eTLD's (effective top-level domains), using the public-suffix list.
The reason is that, for instance, blogspot is so popular that it is pragmatic to think of .blogspot.com similar as to what we'd consider other TLD's, hence the name eTLD.

we do eTLD+1 in Firefox for [tiles currently](https://dxr.mozilla.org/mozilla-central/source/browser/modules/DirectoryLinksProvider.jsm#652).

It goes something like this:
```javascript
Components.utils.import("resource://gre/modules/Services.jsm");
let url;

url = "https://i.can.haz.cheezburger.com";
console.log(Services.eTLD.getBaseDomain(Services.io.newURI(url, null, null)));
// >>> "cheezburger.com"

url = "https://my.awesome.recipes.blogspot.co.uk";
console.log(Services.eTLD.getBaseDomain(Services.io.newURI(url, null, null)));
// >>> "recipes.blogspot.co.uk"
```

Note: `getBaseDomain` illustrated above takes an optional parameter, `aAdditionalParts` which sets the eTLD+aAdditionalParts. The default value is 1.

More:
* https://en.wikipedia.org/wiki/Public_Suffix_List
* https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIEffectiveTLDService#getBaseDomain%28%29

ncloudioj:
Let's just stick to the suggestion from @rlr, using the FQDN for now. Also, i've added an example in the source code to clarify this.

oyiptong:
https://pypi.python.org/pypi/tld seems to do the right thing... but the name is atrocious, as it's very misleading.

This package is appropriately named "publicsuffix": https://pypi.python.org/pypi/publicsuffix
But the function name is misleading. In this case `get_public_suffix` gets the eTLD+1, but the function name says it gets the public suffix.

to be clear, public suffix examples are: com, blogspot.com, co.uk, uk, wordpress.com
eTLD+1 examples are: example.com, nanjiang.blogspot.com

rlr:
ah! interesting... and complicated.

what's wrong with travis?

oyiptong:
Looks like flake8  bails with status code 1

oyiptong:
./splice/queries/distribution.py:152:82: F812 list comprehension redefines 'key' from line 117

tspurway:
r+

Issue #241, opened 2015-11-18
Campaign Cancel doesn't work


oyiptong:
Hitting cancel in the campaign menu doesn't do anything

oyiptong:
There's a console warning:

```
Warning: You cannot PUSH the same path using hash history
```

Probably a pushState bug

k88hudson:
@tkiethanom this is resolved right?

tkiethanom:
@k88hudson I could never reproduce it, and I don't have access to the dev environment. Ricky mentioned the dev environment hasn't been updated.

rlr:
Just tested the cancel button on dev for both the campaign add and edit screens. They both worked. Closing.

PR #245, opened 2015-11-18
Use history 1.13.1 due to hash history bug

k88hudson:
There's a bug that got fixed in https://github.com/rackt/history/commit/adf852fcfc6019d45056c72c90aa4d5f4877c1e8 that is only available in v1.13.1 that affects hash history stuff; I'm pretty sure this will fix it for now


k88hudson:
@tkiethanom r?

tkiethanom:
Does this fix #241 ?

PR #248, opened 2015-11-19
Implement distribution viewer by date using API.

rlr:
r? @ncloudioj 

To test:

```
$ fab build_ui
$ manage.py runserver
$ open http://127.0.0.1:5000/distribution/
```


ui/src/actions/WebtilesPreviewer.js
```diff
@@ -29,97 +36,68 @@ export function selectType(tileType) {
...
 
-function shouldFetchTiles(state, channel, locale) {
-  const localeObj = state.channels[channel].locales[locale];
-  if (localeObj && !localeObj.directoryTiles && !localeObj.isFetching) {
-    return true;
-  }
-  return false;
-}
+            // Get the artifact key
+            let agUrl = index[locale].ag;
```
ncloudioj: The new distribution API will return a list of URLs both for the `ag` and `legacy`. Is this only for the current tile index now?

rlr: Yeah. When you land an updated API, I'll fix that shortly after. Sounds good?

ncloudioj: Sounds good!

I've already merged that PR to sp2 branch. Watch out! :smirk_cat:

rlr: oh... Well let me just add a check here and for now will only use the first item in array. then will fix for real it in a new PR.

rlr:
fixed that ^

ui/src/actions/WebtilesPreviewer.js
ncloudioj: :laughing: oops, short notice.

My bad. Amigo.

ncloudioj:
Once coveralls gets the job done r+!

PR #251, opened 2015-11-20
Integrating zenko ui at /reporting

k88hudson:


k88hudson:
@rlr this is still a bit rough but mergeable I think. do you think you could r?

rlr:
Awesome! I'll start going through it. Travis isn't happy right now.

front_end/package.json
```diff
@@ -30,6 +30,8 @@
     "react-mixin": "3.0.1",
     "react-redux": "4.0.0",
     "react-router": "1.0.0",
+    "react-select": "0.9.1",
+    "react-table": "k88hudson/react-table",
```
rlr: does that mean it grabs your master branch? Just curious.

front_end/src/pages/Reporting/ReportingPage.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { pageVisit } from 'actions/App/AppActions';
+import Nav from './ReportingNav';
+
+const navItems = [
+    {link: '/reporting', icon: 'fa-file-o', label: 'Reports'}
```
rlr: NIT: In the other files you have 2-space indentation. In this one 4.

rlr:
I notice a lot of usage of `.setState(...)` which I hadn't used at all with redux and don't see much of it in the examples. Usually the components are stateless and they get everything they need injected as props (from the redux store). So, I'm not sure if you did it for a reason or that's how you are used to doing things in React without redux or maybe something else.

Otherwise things look good. I haven't fired it up yet.

k88hudson:
@rlr personally I don't think it makes a lot of sense to use redux for state that is local to a single component; obviously for app-level data from the API redux is awesome but it feels like way to much overhead for something like hiding/showing an editor component. happy to discuss if you want

front_end/package.json
k88hudson: yes it does -- generally it's a good idea to lock to a specific commit i.e.

```
k88hudson/react-table#commithash
```

rlr:
Yeah, that seems to agree with what I found when looking into it https://www.reddit.com/r/reactjs/comments/3ogz52/using_setstate_with_redux/

Does hot reloading still work the way you have it? Is the state preserved?

tkiethanom:
That's an interesting idea. I never thought to use setstate in addition to redux. Makes me want to refactor some of my code.

k88hudson:
@rlr since redux uses props to inject state, it shouldn't interact with a component's internal state (they use different mechanisms). Yes, hot reloading still works :)

rlr:
awesome! I'm good with this. I don't have any meaningful local data to test this out but that's what prod is for... Kidding, dev :)

Plus, it doesn't have to be perfect this round.

r+

rlr:
oh, but travis still unhappy. r+ once it is happy

k88hudson:
@rlr I added some fixture data and a script to  `tests/populate_database.py`, so if you run it like `python populate_database.py mozsplice_campaigns mozsplice` where those two arguments are your campaigns and stats db names, it should generate some data

k88hudson:
Thanks for taking a look! :dancer:

k88hudson:
merged in a8154f91666f2516ae4a7138caf76efa1ac68006

PR #253, opened 2015-11-20
Separate dev dependencies and dependencies

k88hudson:
I think this would be a good idea so we can get an idea of what is actually going to be included in our front end js/css bundle.


k88hudson:
what do you think @tkiethanom @rlr?

k88hudson:
I also locked dependencies to specific versions to prevent unexpected bugs from happening on patch version updates. We can add greenkeeper or something to help with keeping them up to date

tkiethanom:
Does the react stuff belong in dependencies?

rlr:
:+1: for doing this

k88hudson:
@tkiethanom yep! good call

tkiethanom:
:+1: looks good to me. Thanks @k88hudson !

PR #254, opened 2015-11-23
Better configuration for front end

k88hudson:
OK, so this does a few things:

* allows for a local settings.conf.js file that's gitignored
* groups all configuration under one global (`__CONFIG__`)
* uses a common file to generate webpack settings for karma/webpack

I wrote the node stuff in node 4 so I need to update travis to use that


tkiethanom:
Cool! Moving the settings file out of the repository is much better. Glad you're making these things more efficient.

k88hudson:
@tkiethanom thanks! what do you think about doing 

```
config.get('FOO')
```
v.s.
```
config.FOO
``

The first style is kinda safer for preventing silent errors because it throws if `FOO` is undefined, but requires a bit of boilerplate

**Comment by tkiethanom** [comment]
@k88hudson Safer sounds better to me, even if there is a little boilerplate.

**Comment by k88hudson** [comment]
Ok all fixed up and rebased. r? @tkiethanom @rlr?

File: front_end/src/actions/AdGroups/AdGroupActions.js [hunk 1]
```diff
@@ -1,11 +1,7 @@
 import fetch from 'isomorphic-fetch';
+import * as config from 'helpers/config';
 
-let apiUrl;
-if (typeof __DEVELOPMENT__ !== 'undefined' && __DEVELOPMENT__ === true) {
-  apiUrl = __DEVAPI__;
-} else {
-  apiUrl = __LIVEAPI__;
-}
+const apiUrl = config.get('DEVELOPMENT') ? config.get('DEV_API') : config.get('LIVE_API');
```
rlr: Now that we have nicer settings. Should we look at just having one `API_URL` setting? Having only DEV_ and LIVE_ doesn't really help given we have local, dev, stage and prod environments.

Doesn't have to go in this PR.

rlr:
I like! r+ from me

front_end/src/actions/AdGroups/AdGroupActions.js
k88hudson: Yeah i agree

k88hudson:
Actually, this might affect our ops situation depending on how we're configuring stuff at the moment and also what version of node we're running, let me just check on that before merging

k88hudson:
@rlr ok i fixed it up so it will work for production as well

rlr:
thank you! r+

PR #259, opened 2015-11-25
Fix #256 Front-end: Add basic error handling for api

### LINKED ISSUE #256: Front-end: Add basic error handling for api
We should design proper error states and some point, but it would be good to at least log errors received in api responses instead of throwing.

tkiethanom:
r? @k88hudson

### DIFF
--- a/front_end/src/actions/Accounts/AccountActions.js
+++ b/front_end/src/actions/Accounts/AccountActions.js
@@ -1,5 +1,6 @@
 import fetch from 'isomorphic-fetch';
 import * as config from 'helpers/config';
+import { fetchHelper } from 'helpers/FetchHelpers';
 
 const apiUrl = config.get('API_URL');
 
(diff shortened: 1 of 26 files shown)


front_end/.eslintrc
```diff
@@ -23,7 +23,8 @@
     "no-duplicate-case": 0,
     "brace-style": 0,
     "func-names": 0,
-    "react/jsx-boolean-value": 0
+    "react/jsx-boolean-value": 0,
+    "no-console": 0
```
k88hudson: Instead of turning off this rule (in general I'm finding it helpful), what do you think about just adding `//eslint-disable-line no-console` to the area you need to do a console log?

In terms of not having to do that every single time... maybe it would be worth abstracting your fetch stuff into a single function you can use in all your actions.

Feel free to file as a follow-up if you need to move on to other things

k88hudson:
Awesome, thanks for doing this! Just a small suggestion, let me know what you think

tkiethanom:
@k88hudson API actions now call a fetchHelper function to handle the fetch. Thanks for the help!

k88hudson:
Thanks :+1: R+

PR #267, opened 2015-11-25
Merge branch 'sp2' into issue-159-image-hashes

tspurway:
re-populates tile.image_uri and tiles.enhanced_image_uri with actual URLs (not hashes)

r? @ncloudioj


ncloudioj:
r+! once fixing the pep8 issues in line 319.

PR #268, opened 2015-11-25
New group_by fields for stats

k88hudson:
Adds `campaign_id`, `account_id`, and `adgroup_id` to list of valid group_by fields in stats api.

E.g.

Request: `/api/stats?group_by=account_id`
Response
```js
{
    "results": [
        {
            "account_id": 1, 
            "blocked": 24, 
            "clicks": 1, 
            "impressions": 6141, 
            "pinned": 0
        }, 
        {
            "account_id": 2, 
            "blocked": 0, 
            "clicks": 0, 
            "impressions": 10, 
            "pinned": 0
        }
    ]
}
```

### DIFF
--- a/splice/queries/reporting.py
+++ b/splice/queries/reporting.py
@@ -1,11 +1,94 @@
 from splice.models import impression_stats_daily
-from splice.queries.tile import get_tiles
+from splice.queries.tile import get_tiles, get_tile_ids_by_group
 from sqlalchemy.sql import func
 from sqlalchemy.orm import aliased
 from sqlalchemy import case
+from splice.queries.common import tuple_to_dict
+from sqlalchemy.exc import InvalidRequestError
+
+CROSS_DB_COLUMNS = {
+    'account_id',
+    'campaign_id',
+    'adgroup_id',
+    'type',
+    'adgroup_type',
+    'category'
+}
+
+CROSS_DB_ERROR = 'You may only group by ONE of %s'
 
 
-def get_stats(group_by, filters=None):
+def build_subquery_table(env, stats_table, cross_db_group_by, group_by, filters=None):
+    """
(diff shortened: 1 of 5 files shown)


k88hudson:
@tkiethanom I decided not to include the full account response in the request since it's a bit tricky to do so across 2 databases, so for now let's just do 2 requests. We can optimize into a single request later if we find it's necessary. Otherwise, it should be working

tkiethanom:
@k88hudson Thanks for the quick turnaround! 

For the record it will be more than 2 requests. 1 request to get the initial stats then 1 request x each account.

k88hudson:
@tkiethanom can't you just use `api/accounts` etc.?

tkiethanom:
@k88hudson You are right, I can get all the accounts that way.

splice/queries/reporting.py
```diff
@@ -3,6 +3,49 @@
...
+    """
+    Build a subquery for group_by fields that require groupings in mozsplice_campaigns
+    """
+    grouped_tiles = get_tiles(group_by=group_by, filters=filters)
+
+    if not grouped_tiles:
+        return None
+    for group in grouped_tiles:
+        for tile in group['tile_ids']:
+            tiles.append(tile)
```
ncloudioj: Mutating directly on the argument is kinda confusing here, unless you do want the side-effect here. Prefer to make a copy. Memory is way cheaper than developer's time :laughing:

k88hudson: oh whoops! good catch

k88hudson:
@ncloudioj ok I fixed up some stuff and refactored into a few helper functions

ncloudioj:
r+! @k88hudson 

Apologies for the delayed r+. Thought I had already dropped down it last week :)

k88hudson:
No worries, thanks @ncloudioj!

PR #275, opened 2015-11-26
Close database sessions in read queries

ncloudioj:


k88hudson:
@ncloudioj LGTM. do we maybe want to merge this since we've been hitting the connection pool error a lot?

tspurway:
Would it make sense if we put these in `try: .... finally:...` blocks?

ncloudioj:
Oops, fugly code :)

It is definitely a dirty try-and-error hack, the flask-sqlalchemy doc says it's unnecessary to explicitly close the session if it is created in the request/response scope. Wanted to try closing every session in case if there is none managed sessions.

I suspect this PR is no longer needed, as the splice log shows that the endpoint `api/init/all` keeps failing and might leave some database sessions behind (my fault, I forgot to include some static file in the MANIFEST.in). I've fixed that, so let's see what happens.

ncloudioj:
Turns out this is no longer needed as we've fixed this issue by other means.

PR #288, opened 2015-12-03
Initial commit of elapsed countries

### LINKED ISSUE #285: UI Gets Clunky When A Campaign Runs in A Lot of Countries
Came across the following situation when a campaign has a long list of target countries. Needed to scroll a long way down to see the other details.

Perhaps we can elapse the country list in this case, and only show a quick preview by default. For instance, "United States, Canada, Germany, ...". Users can expand the list if they want to view the whole list.

screen shot 2015-12-03 at 2 22 21 pm

tkiethanom:
r? @rlr 

Countries will now be hidden behind `...`

Fixes #285


rlr:
Looks good. r+!

PR #294, opened 2016-01-15
Fix #293 - sorting tiles in a descending order by the created time stamp

### LINKED ISSUE #293: Ording Directory Tiles by The Created Datetime Decendingly
Based on the recent feedback, it would make more sense to sort the directory tiles based on the created time stamp descendingly so that newer tiles would be placed to the top-left slots.

ncloudioj:


tspurway:
@ncloudioj can we add another column, maybe called `position_priority` that can be `high = 0`, `medium = 1`, `low = 1`, where the default should be `medium`  

This would allow for the case where we want to impose partial order on the directory tiles.

ncloudioj:
Sure, makes sense.

If the `position_priority` has three options as `high = 2`, `medium = 1`, and `low = 0`, so directory tiles will be sorted by two keys in the descending order, i.e. (position_priority, created_at).

Will add this feature in a separate PR as it requires both UI and server side change.

ncloudioj:
Created a new issue for `position_priority` in #296
 
@tspurway can you give another r? for this PR? Will be working on #296 based on this. Thanks!

tspurway:
r+

PR #299, opened 2016-01-20
Move sass to separate compile step

k88hudson:


k88hudson:
@rlr what do you think about this? I moved all the sass from being inside webpack to be a separate process; it should be easier to debug since stuff is not being injected into style tags anymore, things don't have to be required in more than place, etc.

rlr:
For some reason, I thought this is what you were doing in the first place. The whole webpack way importing styles was confusing to me and it tightly couples the html and the styles which used to be a bad thing but things have changed. I think one of the advantages is to modularize the styles but this page is simple enough where that isn't a big deal.

If we change some styles during dev, do we need to run `npm build:css` again and reload the page? Or did I miss something? Probably not a huge deal, just wondering.

front_end/package.json
```diff
@@ -93,7 +96,10 @@
   },
   "scripts": {
     "test": "karma start --reporters dots --single-run --browsers Firefox && mocha ./node_tests/**/*.test.js",
+    "build:css": "node-sass ./src/main.scss -o dist --importer build_utils/sass-importer.js --source-map dist/main.css.map",
     "build:dist": "gulp build:dist",
-    "start": "gulp serve"
+    "build:static": "cp -r src/static/* dist",
+    "build:fonts": "cp -r node_modules/font-awesome/fonts dist",
+    "start": "mkdir -p dist && npm run build:static && npm run build:fonts && npm run build:css && npm run build:css -- -w & gulp serve"
```
rlr: `build:css` is here twice

k88hudson: you actually need to run it twice because the watch task from node-sass doesn't do an initial build :(

k88hudson:
@rlr the `npm start` task should be sufficient to do everything needed

rlr:
doh! just realized this is splice. No wonder I was a little confused. LGTM

k88hudson:
I will fix the test errors in a new patch, since they're unrelated (babel 5 needs an upgrade)

PR #305, opened 2016-01-25
Fix for build ui script on stage/prod.

rlr:
r? @ncloudioj


ncloudioj:
r+
I think coveralls was unhappy due to the ubiquitous Monday Syndrome. This PR hasn't touched the code base at all.

PR #309, opened 2016-01-26
Fix the invalid image uri

ncloudioj:
Fixed two more bugs found during the migration.
* add campaign for "Get Smart On Surveillance"
* handle the special case where the image_uri is image hash instead of the actual uri


ncloudioj:
@rlr r?

rlr:
coveralls still bust. but wfm. r+

PR #312, opened 2016-02-02
Fix the tile creation to handle the missing images

ncloudioj:
Fix the tile creating function, which should consider that both the imageURI and enhancedImageURI fields are optional in the tile schema.

r? @relud


relud:
lgtm, but I don't know how the client will respond. if you're sure the client is fine with those keys being missing, r+

ncloudioj:
@relud yes, I am pretty sure Firefox will take care of this. The old version of Splice handled this in the same way.

PR #313, opened 2016-02-17
Content signing

ncloudioj:
This implements the content signing. For more detail see `content-signing.md`

The commit history was kinda mixed up as I pull the latest change from master, should be fixed once I merge it back to master.


splice/web/api/content.py
```diff
@@ -0,0 +1,98 @@
...
+    """Signed and upload a piece of content to S3, return the URL if succeeds."""
+    def _is_allowed_file(name):
+        return '.' in name and name.rsplit('.', 1)[1].lower() in set(['zip'])
+
+    content_file = request.files["content"]
+    name = request.args.get('name')
+    version = request.args.get('version')
+
+    if not _is_allowed_file(content_file.filename):
+        return jsonify(message="Invalid content uploaded, expecting a zip file."), 400
```
oyiptong: We should probably make note of this for @mostlygeek.
The nginx `client_max_body_size` should be set sufficiently high.

splice/web/api/content.py
```diff
@@ -0,0 +1,98 @@
...
+            failed.append("%s: %s" % (content['name'], e))
+        else:
+            succeeded.append(content['name'])
+
+    return jsonify(succeeded=succeeded, failed=failed)
+
+
+@content_bp.route('/content/sign', methods=['POST'])
+def handler_content_upload():
+    """Signed and upload a piece of content to S3, return the URL if succeeds."""
```
oyiptong: typo: Sign and upload...

oyiptong: return the URL upon success

splice/web/api/content.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # try to fetch the content from the database
+    new_content = False
+    freeze = False
+    content_record = get_content(name)
+    if content_record is None:
+        content_record = {"name": name}
+        new_content = True
+
+    # determin the version based on the query string (if specified), and the current version in the database
```
oyiptong: typo: determine

splice/web/api/content.py
```diff
@@ -0,0 +1,98 @@
...
+    if version:
+        try:
+            version = int(version)
```
oyiptong: what is the purpose of a query string parameter? It seems the user needs to know in advance what version the content will be

splice/web/api/content.py
```diff
@@ -0,0 +1,98 @@
...
+            if not new_content and content_record["version"] < version:
+                return jsonify(message="Given version %d is ahead of the controlled version %d." % (version, content_record["version"])), 400
+            elif new_content and version != 0:
+                return jsonify(message="Failed to find the content with the given version %d. You can omit the version to create a new content." % version), 400
+
+            if not new_content and version <= content_record["version"]:
+                freeze = True  # re-sign an existing version, do not bump up the version
+    else:
+        if new_content:
+            version = 0
```
oyiptong: start at version 1?

splice/web/api/content.py
```diff
@@ -0,0 +1,98 @@
...
+            succeeded.append(content['name'])
+
+    return jsonify(succeeded=succeeded, failed=failed)
+
+
+@content_bp.route('/content/sign', methods=['POST'])
+def handler_content_upload():
+    """Signed and upload a piece of content to S3, return the URL if succeeds."""
+    def _is_allowed_file(name):
+        return '.' in name and name.rsplit('.', 1)[1].lower() in set(['zip'])
```
oyiptong: instead of checking the filename, how about using libmagic?

It is already included in splice

oyiptong: Addons are shipped with the `xpi` filename, but they are zip files.

splice/web/api/content.py
```diff
@@ -0,0 +1,98 @@
...
+            content_record['version'] = version
+        else:
+            version = content_record['version']
+
+    # sign the content then upload it to S3
+    try:
+        # TODO(najiang@mozilla.com): return the signing pub key and store it to database?
+        urls, new_version = upload_signed_content(content_file.stream, name, version, freeze)
+    except Exception as e:
+        return jsonify(message="%s" % e), 400
```
oyiptong: The error may also be a 50X family, e.g. if an S3 upload failed.
It would be good to catch those. Also:

`except Exception as e` does not catch all exceptions in python 2.7. not all exceptions inherit from `Exception`.

splice/web/api/content.py
ncloudioj: This is to re-sign a specific version, e.g. v1 gets some bug fixes to deploy in an a/b test. The user wants to keep the old version to finish the a/b test. Make sense?

splice/web/api/content.py
oyiptong: Other notable exception: disk full (we're writing the temporary file to disk)

splice/web/api/content.py
ncloudioj: Good point, will bring it up in the meeting.

splice/web/api/content_upload.py
```diff
@@ -0,0 +1,209 @@
...
+        original.close()
+        return None
+    else:
+        return original
+
+
+def resign_content(name, version):
+    """Re-sign a content till the given version. It downloads the original
+    content file from S3, and only re-signs the assets in the manifest. Also,
+    it won't change the version of the content.
```
oyiptong: It *will* change the version of the content

splice/web/api/content_upload.py
```diff
@@ -0,0 +1,209 @@
...
+
+    Params:
+        name: name of the target content
+        version: the current version of the target content
+    Returns: None
+    Exception: an exception will be raised if any error happens
```
oyiptong: That's a tautology :-)

splice/web/api/content.py
ncloudioj: Yup, let's try libmagic then.

For the xpi, since it's already zipped as a single archive, how should we embed the manifest.json. I naively thought we might have to sign it as a single file.

splice/web/api/content_upload.py
```diff
@@ -0,0 +1,209 @@
...
+
+def upload_content_to_s3(name, version, asset, bucket, headers):  # pragma: no cover
+    headers = dict(headers)  # clone the headers as it's mutable
+    asset_name, asset_body, signature = asset
+    s3_key = "{0}/v{1}/{2}".format(name, version, asset_name)
+
+    key = Key(bucket)
+    key.name = s3_key
+    _, ext = os.path.splitext(asset_name)
+    headers['Content-Type'] = MIME_EXTENSIONS.get(ext) or 'text/plain'
```
oyiptong: This is good.

We can't use libmagic here, because libmagic returns the wrong MIME type for xpi and css.

splice/web/api/s3_common.py
```diff
@@ -0,0 +1,27 @@
...
+    from splice.environment import Environment
+    from boto.s3.cors import CORSConfiguration
+
+    env = Environment.instance()
+    bucket = env.s3.get_bucket(env.config.S3[bucket])
+    cors = CORSConfiguration()
+    cors.add_rule("GET", "*", allowed_header="*")
+    bucket.set_cors(cors)
+    headers = {
+        'Cache-Control': 'public, max-age=31536000',
```
oyiptong: We actually don't want that max-age for html pages. This may be fine for other resources.

Can we set a max-age to 24h for html files?

splice/web/api/content.py
```diff
@@ -0,0 +1,98 @@
...
+        except Exception as e:
+            failed.append("%s: %s" % (content['name'], e))
+        else:
+            succeeded.append(content['name'])
+
+    return jsonify(succeeded=succeeded, failed=failed)
+
+
+@content_bp.route('/content/sign', methods=['POST'])
+def handler_content_upload():
```
oyiptong: We need to ensure nginx will not time out. This endpoint will take a while.

oyiptong: The current configs inherited from the past incarnation of Splice are probably OK. Just need to make note of this, maybe in a comment.

splice/web/api/content_upload.py
ncloudioj: Hmm, it's kind of confusing here.

Should we consider re-signing the content in-place for all the existing version? Say we've already had 3 different versions for a content. They're all being used by Ffx users. If the signing key gets compromised. Shouldn't we just re-sign those 3 in-place?

splice/models.py
```diff
@@ -6,6 +6,16 @@
 metadata = db.metadata
 
 
+class Content(db.Model):
+    __tablename__ = "contents"
+
+    id = db.Column('id', db.Integer(), autoincrement=True, primary_key=True, info={"identity": [1, 1]})
+    name = db.Column('name', db.String(length=255), nullable=False, unique=True)
+    version = db.Column('version', db.Integer(), nullable=False)
+    created_at = db.Column('created_at', db.DateTime(), server_default=db.func.now(), nullable=False)
```
oyiptong: Might want to store original content bucket/keyname here. I know it's generated in `_get_original_content`, but the key name structure may change or we may move buckets.

splice/web/api/content_upload.py
ncloudioj: Heh, alright, you caught me :cry:

splice/web/api/s3_common.py
ncloudioj: Sure will do.

splice/models.py
oyiptong: You also want to store the digest for the original zip file. Additionally, upon re-signing, we need to verify that the digest has not changed.

oyiptong: If not, this leaves us to an attack vector:

1. CDN owner is compromised, changes original zip file as well as HTML files
2. We discover the issue, regenerate keys and re-sign
3. We sign the compromised original zip file

splice/web/api/content.py
ncloudioj: Yes, I think the current nginx timeout is 10min, kudos to suggested tiles. We should be OK for a while.

splice/default_settings.py
```diff
@@ -41,11 +41,16 @@ class DefaultConfig(object):
 
     S3 = {
         "bucket": "moz-tiles-dev",
+        "content": "moz-content-dev",
```
oyiptong: might want a separate bucket for original content. Added a note in model.py review

splice/web/api/content_upload.py
```diff
@@ -0,0 +1,209 @@
...
+                    continue
+                upload_content_to_s3(name, v, asset, bucket, headers)
+        except:
+            raise
+        finally:
+            original.close()
+
+
+def upload_signed_content(content, name, version, freeze=False):
+    """Upload the signed content file to S3, return all the urls if succeeds
```
oyiptong: This also uploads the original content to S3. We need a better function name and descriptive comments.

splice/models.py
ncloudioj: Right on! Very good points.

splice/web/api/content.py
oyiptong: We should be covered, again due to our big ingest file sizes ;-)

splice/web/api/content.py
oyiptong: Ah! got it. Awesome

splice/web/api/content_upload.py
oyiptong: Hmm. I think you are right. We will have an offline backup.

The signature verification will fail and we'll have them use it.

oyiptong:
Overall, looking good. Needs a couple more changes.
f+

oyiptong:
(where f+ means feedback+)

splice/web/api/content_upload.py
ncloudioj: As a side note, in the current implementation, the signature contains the x5u, which points to the public signing key. I think the Content Verifier in Firefox is looking to use PKI eventually instead of using the hard-coded keys. See more detail [here](https://github.com/mozilla-services/autograph#response)

ncloudioj:
Thank you @oyiptong for teaching me yet another *F* word!

Those are all very insightful points, will be fixing/adding them to the code base shortly.

oyiptong:
as per @mostlygeek's comments, we'll need to turn off tiles upload UI and endpoints. Maybe we want to do this via a config?

At some point we should make this its own server.

mostlygeek:
maybe you should fork this and delete that unnecessary tile UI/endpoints. :)

oyiptong:
You may have a point. It does take less work to add the config to drop routes, however, it will be more work later on to extricate the two code bases.

ncloudioj:
Add various things based on the feedback:

1. verify signatures generated by the signing server
2. add a `versions` table to track the `original_url` and `original_hash`. Content integrity will be checked when user requests for re-sign-all
3. original content will be retrieved from S3 first. Upon failure, it will retry the downloading from the `original_url` directly

@oyiptong One more run?

oyiptong:
removing some comments addressed in a later commit

splice/web/api/content_upload.py
```diff
@@ -0,0 +1,274 @@
...
+        original.close()
+        return None
+    else:
+        return original
+
+
+def _check_content_integrity(content_file, checksum):
+    content_file.seek(0)
+    raw = content_file.read()
+    return base64.b64encode(hashlib.sha1(raw).digest()) == checksum
```
oyiptong: `hashlib.sha1(raw).hexdigest()` will work better

splice/web/api/content_upload.py
```diff
@@ -0,0 +1,274 @@
...
+    urls.sort()
+
+    # upload the original content to its bucket
+    bucket_original, _ = setup_s3(bucket="content-original")
+    content.seek(0)  # rewind the file offset to read the whole content
+    raw = content.read()
+    asset = (_ORIGINAL_NAME, raw, None)
+    url = upload_content_to_s3(name, version, asset, bucket_original, headers)
+    urls.append(url)
+    original_hash = base64.b64encode(hashlib.sha1(raw).digest())
```
oyiptong: `hashlib.sha1(raw).hexdigest()` will work better

oyiptong:
@ncloudioj feedback+ we're almost there!

oyiptong:
1. how about storing `hashlib.sha1(raw).hexdigest()` in the version `original_hash` column? it would be shorter than the base64 encoded one.
2. max-age of 24h only applies to HTML files, the other files should be cached for a long time

oyiptong:
r+ after fixes!

ncloudioj:
Yup, that sounds reasonable.

Also will change the caching policy as you mentioned here. Thanks for
pointing it out!

Issue #317, opened 2016-02-26
Add config to expose content-signing related endpoints only


ncloudioj:
When deploying Splice as a content signing application, it makes sense to disable other signing irrelevant services (tile uploading, reporting, etc.). This feature aims to add the configurability for Splice to do so.

PR #318, opened 2016-02-26
Add configuration to expose content signing only

ncloudioj:
This implements feature for #317 

r? @oyiptong

### DIFF
--- a/splice/default_settings.py
+++ b/splice/default_settings.py
@@ -10,6 +10,7 @@ class DefaultConfig(object):
     """
     DEBUG = True
     APPLICATION_ROOT = None
+    CONTENT_SIGNING_ONLY = False
     JSONIFY_PRETTYPRINT_REGULAR = True
 
     STATIC_ENABLED_ENVS = {"dev", "test"}
--- a/splice/webapp.py
+++ b/splice/webapp.py
@@ -5,8 +5,13 @@ from splice.environment import Environment
 register_flask_restful = False
 
 
-def setup_routes(app):
+def setup_routes(app, content_signing_only):
     global register_flask_restful
+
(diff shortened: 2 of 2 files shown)


splice/webapp.py
```diff
@@ -5,9 +5,14 @@
 register_flask_restful = False
 
 
-def setup_routes(app):
+def setup_routes(app, content_signing_only):
```
oyiptong: you don't need this, you can do `env.application.config`

splice/default_settings.py
```diff
@@ -10,6 +10,7 @@ class DefaultConfig(object):
     """
     DEBUG = True
     APPLICATION_ROOT = None
+    CONTENT_SIGNING_ONLY = False
```
oyiptong: Might be better to have ALLOWED_APPS = ['tiles', 'signing']

splice/webapp.py
```diff
@@ -5,9 +5,14 @@
...
     global register_flask_restful
 
+    import splice.web.api.content
```
oyiptong: better to put a if statement for each type of API endpoint being registered.

e.g.
```python
if 'tiles' in app.config.ALLOWED_APPS:
    """do tiles API registration"""

if 'signing' in app.config.ALLOWED_APPS:
    import splice.web.api.content
    splice.web.api.content.register_routes(app)
```

Much better than short-circuiting the function

oyiptong:
feedback+

splice/webapp.py
ncloudioj: Yes, that looks way better.

ncloudioj:
Landed in cce7f46f02041ecda82b1fd531baf5618038fa9d

PR #323, opened 2016-03-02
Add hawk_id to the config file

ncloudioj:
Also upgrade gevent to 1.1rc5 as we found a ssl issue with gevent in stage.


relud:
r+

PR #325, opened 2016-03-04
Fix the base64url padding issue

ncloudioj:


ncloudioj:
@jvehent one more r?

Keep it in sync with upstream :)

jvehent:
The `r+` you shall haz!

Issue #27, opened 2014-10-28
Implement batched pings for Fennec


oyiptong:
We will be collecting batched pings from Fennec.

This implements the data model and reporting.

[Data collection](https://github.com/oyiptong/onyx/issues/10) in onyx
[Aggregation](https://github.com/tspurway/infernyx/issues/7) in infernyx

PR #328, opened 2016-03-10
Alter activity_stream_stats_daily table

ncloudioj:
Based on [#27](https://github.com/mozilla/activity-streams/issues/27), we have tweaked the schema of 'activity_stream_stats_daily'.

Since it's still in stage, we made this change in place instead of creating a new migration.


ncloudioj:
A few tweaks on the table schema for Activity Stream.

@oyiptong r?

oyiptong:
r+

PR #330, opened 2016-03-11
Issue 329 adapt upstream changes

ncloudioj:
Added two changes to catch up with Autograph:
- update the signing service endpoint from `signature` to `sign/hash`
- prepend a template for each raw content prior to hashing it

Also, we've added the bookkeeping for the signing public key, as Autograph always includes it in the signature payload.

### DIFF
--- a/splice/web/api/content.py
+++ b/splice/web/api/content.py
@@ -29,7 +29,11 @@ def handler_content_resign_all():
     succeeded, failed = [], []
     for content in get_contents():
         try:
-            resign_content(content)
+            new_pub_key = resign_content(content)
+            update_record = {"signing_key": new_pub_key}
+            with session_scope() as session:
+                for version in content["versions"]:
+                    update_version(session, content["id"], version["version"], update_record)
         except Exception as e:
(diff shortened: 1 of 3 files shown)


ncloudioj:
@oyiptong r?

As I mentioned yesterday :grin:

splice/web/api/content.py
```diff
@@ -29,7 +29,11 @@ def handler_content_resign_all():
     succeeded, failed = [], []
     for content in get_contents():
         try:
-            resign_content(content)
+            new_pub_key = resign_content(content)
+            update_record = {"signing_key": new_pub_key}
```
oyiptong, line 33: do you mean `new_record`?

using a verb for a data structure is confusing

oyiptong, line 33: perhaps you meant `updated_record`?

oyiptong:
r+ with nits

splice/web/api/content.py
ncloudioj, line 33: yes, it should be `updated_record`

Issue #333, opened 2016-03-18
Better error message for invalid manifest.json 


ncloudioj:
The current message is not clear on reporting what the error is, for example, if the `bump_version` is missing in `manifest.json`, the error is

``` json
{
  "message": "bump_version"
}
```

Issue #332, opened 2016-03-18
default version number should be 1 for the signing API


oyiptong:
I just uploaded a binary for the first time and it set the default version to zero.

```
$ http -f POST http://splice_signing_tiles.stage.mozaws.net/api/content/sign\?name\=newtab content@./signing-test-content.zip
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 1388
Content-Type: application/json
Date: Fri, 18 Mar 2016 15:15:52 GMT

{
    "content": {
        "created_at": "Fri, 18 Mar 2016 15:15:48 GMT",
        "id": 2,
        "name": "newtab",
        "version": 0
    },
    "uploaded": [
        "https://net-mozaws-stage-splice-content.s3.amazonaws.com/newtab/v0/aurora/en-US/index.html",
        "https://net-mozaws-stage-splice-content.s3.amazonaws.com/newtab/v0/beta/en-US/index.html",
        "https://net-mozaws-stage-splice-content.s3.amazonaws.com/newtab/v0/esr/en-US/index.html",
        "https://net-mozaws-stage-splice-content.s3.amazonaws.com/newtab/v0/manifest.json",
        "https://net-mozaws-stage-splice-content.s3.amazonaws.com/newtab/v0/nightly/en-US/index.html",
        "https://net-mozaws-stage-splice-content.s3.amazonaws.com/newtab/v0/release/en-US/index.html",
        "https://net-mozaws-stage-splice-content-original.s3.amazonaws.com/newtab/v0/original.zip"
    ],
    "version": [
        {
            "content_id": 2,
            "created_at": "Fri, 18 Mar 2016 15:15:48 GMT",
            "last_updated": "Fri, 18 Mar 2016 15:15:48 GMT",
            "original_hash": "fc06da43c65e701759a2cb41289dc7e316a9508073aa20602e27ab6bd478cea4",
            "original_url": "https://net-mozaws-stage-splice-content-original.s3.amazonaws.com/newtab/v0/original.zip",
            "signing_key": "MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAEC9SPnO64nyU9ZvtJ2iYiifxdQOZhcrNVvzA5Z8TbRvhr3e9YGdBNsexwx/P90x2+ZNHuIK7h4riyMGU57GW+zmGRNOu35aC0qsRxs3REsgKON+XyShHoeivuIjOLaloP",
            "version": 0
        }
    ]
}
```

PR #334, opened 2016-03-18
Hotfixes for issue 332 and issue 333

ncloudioj:
This has two fixes for #332 and #333, respectively. Found during the end-to-end signing test.


oyiptong:
r+

PR #338, opened 2016-04-05
update alembic migration to respect ignore redshift option

relud:
r? @ncloudioj


ncloudioj:
r+

relud:
this doesn't need its own release, imo

ncloudioj:
How about I merge and tag it as 2.1.7.1 for the upcoming production deploy?

relud:
_shrug_ anything works for me, including just deploying 2.1.7 as it was to prod.

ncloudioj:
ah, cool, thought it might incur some errors during the deploy without this guard.

k, merging it without bumping the version. Thanks.

PR #340, opened 2016-04-18
Add script to expire redshift data

ncloudioj:
This adds a new script which generates SQL statements to expire data in the Redshift database.

It purges the table through the "delete-vacuum-analyze" cycle. Also note that we need to manually execute the SQL statements against the server after reviewing them carefully.

### DIFF
--- a/scripts/expire_redshift_data.py
+++ b/scripts/expire_redshift_data.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python
+from optparse import OptionParser
+from datetime import date, timedelta
+
+
+USAGE = """Usage:
+        # delete all the rows that are more than 7 days old
+        %prog -n 7
+
+        # delete with specific tables
+        %prog -n 30 -t the_table_name
+        """
+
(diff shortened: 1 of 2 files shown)


relud:
lgtm

scripts/expire_redshift_data.py
```diff
@@ -0,0 +1,53 @@
...
+_TABLES = ["impression_stats_daily", "application_stats_daily",
+           "newtab_stats_daily", "site_stats_daily"]
+_TEMPLATE = "DELETE FROM {table} WHERE date < {date_lower};"
+_TEMPLATE_VACUUM = "VACUUM {table};"
+_TEMPLATE_ANALYZE = "ANALYZE {table};"
+
+
+def main():
+    parser = OptionParser(usage=USAGE)
+    parser.add_option("-n", "--days",
```
tspurway, line 24: This option is dangerous.  Let's remove the '-n' optional altogether and change '--days' to '--days-to-retain'

PR #349, opened 2016-05-26
Add experiment_id for activity_stream_events_daily

ncloudioj:


tspurway:
@ncloudioj - are we going to use the `experiment_id` in the `activity_stream_stats_daily`?  This isn't a request, I was just wondering what your thoughts are on that.

ncloudioj:
Not for this experiment. But good point, we can add it to all the AS tables for the upcoming experiments. Will make the change for Infernyx as well.

PR #352, opened 2016-06-01
update requirements.txt

relud:
builds are failing without these requirements

r? @ncloudioj


relud:
I know the setuptools requirement is something that changed out from under us, because it's happened to a bunch of other projects too. I don't understand why pycrypto is suddenly needed, but this fixes it.

relud:
that build failure doesn't look related to python deps. looks like something node related.

ncloudioj:
r+

PR #354, opened 2016-07-22
Fix the case that resigning an exsiting content without bumping version

ncloudioj:
This fixes the issue #353


ncloudioj:
@oyiptong r?

PR #372, opened 2017-02-14
fix(ping-centre): Add table for test pilot

### LINKED ISSUE #371: Add table for the Test Pilot topic in Ping-centre

ncloudioj:
This closes #371.


ncloudioj:
r? @emtwo or @6a68 

A few things to note here,

* Both 'receive_at' and 'date' were not defined in the original schema, i added those two to the table schema as they would be important for the dashboard. Note that the client doesn't need to include them in the ping as they'd be populated by Onyx
* The size of `raw` field is limited by 16k bytes. In theory, Redshift supports up to 65535 bytes for the string type. However, using such a big string field is not recommended as it could grow up the database quickly. That being said, if the `raw` field is bigger than 16k, the ETL job will drop this record. Does 16k sound a reasonable size for your use case? @6a68

ncloudioj:
@emtwo Could you r? this one, please? I've also added the table schema for A-S mobile so that we can just do a single deploy for both T-P and A-S.

emtwo (approved):
Looks good!

PR #424, opened 2018-03-08
Update signing protocol for autograph

ncloudioj:
This adapts the new signing protocol of [Autograph](https://github.com/mozilla-services/autograph)

* Add `keyid` to the signing request in order to let Autograpy generate `x5u` field in the response. See more detail [here](https://github.com/mozilla-services/autograph/blob/master/docs/endpoints.rst#11request)
* Extract `x5u` from the response so that it can be used when verifying the signature in Firefox. See more detail [here](https://github.com/mozilla-services/autograph/blob/master/docs/endpoints.rst#12response)


rlr (approved):
lgtm 🚢

PR #426, opened 2018-05-22
Add table for activity stream router

### LINKED ISSUE #425: Add table for AS router
This is for the Activity Stream Router Experiments.

ncloudioj:
This closes #425 

r? @jbuck Could you take a look at it please? Particularly the encoding and distkey&sortkey part.


ddl/migrations/redshift/1526998293.sql
```diff
@@ -0,0 +1,24 @@
...
+    date DATE NOT NULL ENCODE ZSTD, 
+    locale VARCHAR(14) NOT NULL ENCODE BYTEDICT, 
+    country_code VARCHAR(5) NOT NULL ENCODE BYTEDICT, 
+    os VARCHAR(64) NOT NULL ENCODE ZSTD, 
+    browser VARCHAR(64) NOT NULL ENCODE ZSTD, 
+    version VARCHAR(64) NOT NULL ENCODE ZSTD, 
+    device VARCHAR(64) NOT NULL ENCODE ZSTD
+)
+DISTKEY (impression_id)
+SORTKEY (event);
```
jbuck: Do you know what kinds of queries will be using this table?

ncloudioj: I'm going to use it most likely :)

SELECT xyz FROM here WHERE event = foo and source = bar

Perhaps I can add source to the sortkey

ncloudioj: Although, as a starter, this will be used in various experiments conducted in the beta channel.

jbuck: Do you think you'll be aggregating over days as well? or just source/event queries?

ncloudioj: Good catch, `date` is always a frequent term in the where clause.

So perhaps `SORTKEY (date, event, source)`?

jbuck (approved):

